Issue #9, opened 2021-09-02
Allow multi user usage


TheSlimvReal:
Currently the controller only manages one logged in user at a time. In order to allow many users to access the app simultaneously this has to be changed. 

This controller should hold the usernames and the associated roles of the logged in users.

Each logged in user is identified by a `uuid` that is set as a cookie after logging in against the `session` endpoint.

Issue #4, opened 2021-09-02
read url and admin password from env


TheSlimvReal:
The URL to CouchDB and the admin password should be passed to the application through environment variables

PR #10, opened 2021-09-02
Authentication

TheSlimvReal:
Includes #9

This PR implements the user authentication.

- [x] Provides the session endpoint to which the user first has to authenticate. On success a JWT token is returned as a cookie. Authentication is currently done against CouchDB but this could be swapped out to another provider.
- [x] All CouchDB replication endpoints require a valid JWT token
- [x] The JWT token holds (encrypted) information about name and roles of the user. This makes the server stateless and allows multiple users to send simultaneous requests.

Missing:

- [x] Currently the JWT token has a fixed lifetime (60s). Under which circumstances should the token be refreshed? @sleidig


src/app.module.ts
```diff
@@ -2,10 +2,13 @@ import { Module } from '@nestjs/common';
...
 
 @Module({
-  imports: [HttpModule],
+  imports: [HttpModule, SessionModule],
   controllers: [CouchProxyController],
   providers: [DocumentFilterService],
 })
 export class AppModule {}
+
+export const COUCH_ENDPOINT = 'https://dev.aam-digital.com/db';
```
sleidig, line 14: Let's use environment variables on the server-side services (probably using dotenv https://nodejs.dev/learn/how-to-read-environment-variables-from-nodejs)

src/couch-proxy/couch-proxy.controller.ts
```diff
@@ -23,44 +24,29 @@ import {
...
-@Controller('couchdb/db')
+@UseGuards(JwtGuard)
+@Controller()
 export class CouchProxyController {
-  readonly couchDB = 'https://dev.aam-digital.com/db';
-  private username: string;
-  private password: string;
-  public userRoles: string[] = ['user'];
+  private username: string = 'demo';
+  private password: string = 'pass';
```
sleidig, line 35: do we still need these at all?

sleidig:
In my opinion the server login should not expire any time while a user is having the app open. So probably at least within 1-2 hours of inactivity users should get an automatically refreshed token? Once the server login expires we should also log out in the local app (and that should rarely happen during a workday, otherwise we'll drive users crazy)

src/app.module.ts
TheSlimvReal: #4

PR #34, opened 2021-11-30
Write -> Create/Update/Delete

sleidig:


src/replication/couch-proxy/couch-proxy.controller.ts
```diff
@@ -223,18 +229,13 @@ export class CouchProxyController {
...
-    console.log('GET bulk doc called', queryParams);
-    return this.httpService
-      .get(`${this.databaseUrl}/${this.databaseName}/_bulk_get`, {
-        params: queryParams,
-      })
-      .pipe(map((response) => response.data));
+  @Get('/:db/_all_docs')
+  allDocsGet(
+    @Query() queryParams: any,
+    @Body() body: AllDocsRequest,
```
TheSlimvReal, line 235: This might be confusing as a `Get` endpoint can't have a body. Either we merge the function with `allDocs` and just assign it to both endpoints or remove the `body` param here and provide a proper dummy value

sleidig: good point, shouldn't have the body parameter here then. Maybe just pass {} or null as body?

sleidig (commented):
Looks very clean and clear :+1:   (can't approve, as I am the original creator of the PR :see_no_evil: )

TheSlimvReal (approved):

PR #90, opened 2022-12-20
feat: public and attachment endpoints

TheSlimvReal:
This PR adds new endpoints to allow uploading and downloading of attachments, based on permissions on the entity to which the attachments are linked.

Additionally, this PR adds a new `public` field to the permission object that allows to define permissions for unauthenticated users.
This can be used for e.g. creating a public registration form.

To achieve this, the `CombinedAuthGuard` passes the authentication even if the login was not possible.

What we should check is how this works together with the frontend. Some possible issues would be users whose token/cookie has expired should correctly receive a 401 if the operation was also not possible with the `public` permissions.


src/auth/guards/combined-auth/combined-auth.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+        if (onlyAuthenticated) {
+          throw err;
+        } else {
+          req.user = undefined;
+          return true;
+        }
+      });
+  }
+
+  use(req: any, res: any, next: () => void) {
```
sleidig, line 61: when is `use` and when `canActivate` relevant? Can you maybe add a short JsDoc to explain that (and explain why for "use" there is no fallback for unauthorized users)?

src/permissions/rules/rules.service.spec.ts
```diff
@@ -141,4 +141,13 @@ describe('RulesService', () => {
...
     expect(() => service.getRulesForUser(normalUser)).toThrow(ReferenceError);
   });
+
+  it("should only return 'public' rules if no user object is passed", () => {
+    testPermission.data.default = [{ subject: 'Config', action: 'read' }];
+    const publicRule: DocumentRule = { subject: 'User', action: 'create' };
+    testPermission.data.public = [publicRule];
+
+    const result = service.getRulesForUser(undefined);
+    expect(result).toEqual([publicRule]);
```
sleidig, line 151: `default` role does not apply to unauthenticated users, right? Maybe we could add an `expect(result)not.toContain(testPermission.data.default);` to make that even more explicit

sleidig (commented):
Should we require an API token for access to public endpoints, if no valid user credentials exist? That would safeguard against attackers that just scan ports - and we could even consider linking a token to a domain for which it is valid

sleidig (approved):

Issue #7, opened 2021-09-02
Re-sync after changed permissions


TheSlimvReal:
Once the permissions change, the sync should be restarted for each client.

TheSlimvReal:
**TODO server side:**
When a user gains access to more document after the permissions are changed, the replication protocol should automatically ensure that the user gets the document synced. This can be achieved by deleting the replication history in the `_local` database.
- [ ] Execute the `clearLocal` endpoint/function whenever the permission rules change.

**TODO client side:**
When a user looses access to documents after the permissions are changed, they should be automatically removed from the clients local database. Currently the only feasible way to achieve this, is to implement it on the client side. When deleting the documents the default `PouchDB` API might not be the right solution because it will create a new revision number that might cause problems in a later sync. Instead the document should be entirely deleted. Therefore a direct access to the `IndexedDB` might be necessary.
- [ ] When permissions change the client should somehow evaluate, to which documents it does not have access any more and completely delete them from the local device.

sleidig:
also see #2300

sleidig:
Somehow the current auto-triggering of clearLocal does not work correctly yet, although it's execution through a manual API call works.

Issue #55, opened 2022-02-18
Automatic detection of changed permission object


TheSlimvReal:
By hooking into the `_changes` feed, the backend should be able to automatically get notifications about updates of the `Permission` entity and update the internal `rules` representation

PR #107, opened 2023-07-26
Automatic permission updates

TheSlimvReal:
See #55 

This addition uses the `_changes` feed to automatically receive updates of the permission object in the backend. This should reduce the risk of forgetting to use the `admin/reload` endpoint after updating the permission object.

### DIFF
--- a/src/permissions/rules/rules.service.spec.ts
+++ b/src/permissions/rules/rules.service.spec.ts
@@ -1,4 +1,4 @@
-import { Test, TestingModule } from '@nestjs/testing';
+import { Test } from '@nestjs/testing';
 import { DocumentRule, RulesService } from './rules.service';
 import { UserInfo } from '../../restricted-endpoints/session/user-auth.dto';
 import { defer, of, throwError } from 'rxjs';
@@ -31,7 +31,7 @@ describe('RulesService', () => {
     } as any;
     jest.spyOn(mockCouchDBService, 'get').mockReturnValue(of(testPermission));
 
-    const module: TestingModule = await Test.createTestingModule({
+    const module = await Test.createTestingModule({
       providers: [
         RulesService,
         {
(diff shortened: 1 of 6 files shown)


src/permissions/rules/rules.service.spec.ts
```diff
@@ -44,33 +53,52 @@ describe('RulesService', () => {
...
   });
 
-  it('should fetch the rules from the db on startup', async () => {
+  it('should fetch the rules from the db on startup', () => {
     expect(mockCouchDBService.get).toHaveBeenCalledWith(
       DATABASE_NAME,
-      Permission.DOC_ID,
+      '_changes',
+      expect.anything(),
     );
```
sleidig, line 68: is this a relevant test? I think the call to CouchDB is not triggering anything important, rather what we may want to check here is whether the ruls in our RulesService have be correctly initialized?

src/permissions/rules/rules.service.ts
```diff
@@ -26,20 +28,40 @@ export class RulesService {
...
+          this.permission = changes.results[0].doc.data;
+        }
+      }),
+      catchError((err) => {
+        console.error('LOAD RULES ERROR:', err);
+        throw err;
+      }),
+      retry({ delay: 1000 }),
+      repeat(),
     );
```
sleidig, line 63: any special reason why to return an observable here that doesn't return things to use in subscribe but in itself repeats once it is subscribed?

Wouldn't it be less convoluted to subscribe to a repeating couchdb request here, move the actual action from `tap` to `subscribe` and not return anything?

src/permissions/rules/rules.service.spec.ts
```diff
@@ -44,33 +53,52 @@ describe('RulesService', () => {
...
+    jest.advanceTimersByTime(30000);
+
+    // 2x error, 1x success, 1x waiting for next change
+    expect(calls).toEqual(4);
+    expect(console.error).toHaveBeenCalledWith(
+      'LOAD RULES ERROR:',
+      expect.any(Error),
+    );
+    expect(console.error).toHaveBeenCalledTimes(2);
+    expect(service.getRulesForUser(normalUser)).toEqual(userRules);
```
sleidig, line 101: maybe one more test that rules are the new rules after a _change event occured?

sleidig (approved):
Code looks good. :+1: 
We should test a deployment to make sure this works seamlessly in practice - haven't done any testing.



is this also a way to resolve #7 later?

src/permissions/rules/rules.service.ts
TheSlimvReal: Good point. Initial had only allowed a maximum amount of errors but this will now repeat forever so the observable doesn't help much.

PR #112, opened 2023-08-16
Restrict CouchDB access

### LINKED ISSUE #82: Permissions on other endpoints that can retrieve docs
Not all endpoints that return documents are currently filtered based on the permissions. E.g. permissions for the `_find` endpoint are still missing. These should also be implemented in order to provide full security.

TheSlimvReal:
closes #82

Based on recent feedback this PR changes the setup to only allow requests that are required by the PouchDB replication algorithm instead of automatically creating a proxy for all others.

Also the `include_docs: true` does not work on the `_changes` feed.

### DIFF
--- a/src/restricted-endpoints/restricted-endpoints.module.ts
+++ b/src/restricted-endpoints/restricted-endpoints.module.ts
@@ -4,7 +4,6 @@ import { DocumentModule } from './document/document.module';
 import { ReplicationModule } from './replication/replication.module';
 import { SessionController } from './session/session.controller';
 import { AuthModule } from '../auth/auth.module';
-import { createProxyMiddleware, RequestHandler } from 'http-proxy-middleware';
 import { CombinedAuthGuard } from '../auth/guards/combined-auth/combined-auth.guard';
 import { AttachmentModule } from './attachment/attachment.module';
 
@@ -13,11 +12,6 @@ import { AttachmentModule } from './attachment/attachment.module';
   controllers: [SessionController],
 })
 export class RestrictedEndpointsModule implements NestModule {
-  /**
-   * This proxy allows to send authenticated requests to the real database
-   */
-  static proxy: RequestHandler;
-
   configure(consumer: MiddlewareConsumer): any {
     consumer
       .apply(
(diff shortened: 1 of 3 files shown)


src/restricted-endpoints/replication/replication-endpoints/replication-endpoints.controller.ts
```diff
@@ -36,6 +37,72 @@ export class ReplicationEndpointsController {
...
+   * Get the changes stream.
+   * The `include_docs` params is automatically set to false.
+   * @param db
+   * @param params
+   */
+  @Get(':db/_changes')
+  changes(@Param('db') db: string, @Query() params) {
+    return this.couchdbService.get(db, '_changes', {
+      ...params,
+      include_docs: false,
```
sleidig, line 102: Instead of overwriting include_docs to always be false, we could consider returning an error code and denying it for include_docs completely. (Client making the request is currently not getting back what they should be able to expect.)

But I like the simplicity of this and don't see a big disadvantage currently :+1:

src/restricted-endpoints/restricted-endpoints.module.ts
```diff
@@ -24,43 +19,14 @@ export class RestrictedEndpointsModule implements NestModule {
...
         urlencoded({ extended: true, limit: '10mb' }),
       )
       .forRoutes('*');
-  }
-
-  private applyProxyForPermissionlessCouchdbEndpoints(
-    consumer: MiddlewareConsumer,
-  ) {
+    consumer.apply(CombinedAuthGuard).exclude('_session');
     this.initializeProxy();
```
sleidig, line 29: do we still need `initializeProxy` at all? RestrictedEndpointsModule.proxy is not applied anymore, so it seems to be unused?

sleidig (commented):
Code looks good :+1:

Do we have a test build that can be deployed to the dev system? I would like to test the sync with this and ensure we didn't accidentally break anything.

src/restricted-endpoints/restricted-endpoints.module.ts
TheSlimvReal: Its still used for the file upload but I can move it there

PR #113, opened 2023-08-21
Changes filter

### LINKED ISSUE #108: Filter documents on changes feed
By proxying the changes feed and using `inculde_docs` true when sending it to CouchDB it would be possible to already filter the documents on the changes feed instead of only when requesting the docs later on.

TheSlimvReal:
Closes #108 

By filtering the document on the changes feed, the network traffic for the replication backend is reduced as less documents will be requested by the frontend. This also even hides the doc ids from the frontend.

### DIFF
--- a/src/restricted-endpoints/replication/changes/changes.controller.spec.ts
+++ b/src/restricted-endpoints/replication/changes/changes.controller.spec.ts
@@ -1,0 +1,129 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { ChangesController } from './changes.controller';
+import { CouchdbService } from '../../../couchdb/couchdb.service';
+import { PermissionService } from '../../../permissions/permission/permission.service';
+import { authGuardMockProviders } from '../../../auth/auth-guard-mock.providers';
+import {
+  ChangeResult,
+  ChangesResponse,
+} from '../bulk-document/couchdb-dtos/changes.dto';
+import { DatabaseDocument } from '../bulk-document/couchdb-dtos/bulk-docs.dto';
+import { UserInfo } from '../../session/user-auth.dto';
+import { firstValueFrom, of } from 'rxjs';
+import {
+  DocumentRule,
+  RulesService,
+} from '../../../permissions/rules/rules.service';
+
+describe('ChangesController', () => {
+  let controller: ChangesController;
+  let mockCouchdbService: CouchdbService;
+  let mockRulesService: RulesService;
+  const schoolDoc: DatabaseDocument = { _id: 'School:1' };
+  const privateSchoolDoc: DatabaseDocument = { _id: 'School:2', private: true };
+  const childDoc: DatabaseDocument = { _id: 'Child:1' };
+  const deletedChildDoc: DatabaseDocument = { _id: 'Child:2', _deleted: true };
+  const changes: ChangesResponse = {
+    last_seq: 'some_seq',
+    results: [schoolDoc, privateSchoolDoc, childDoc, deletedChildDoc].map(
+      docToChange,
+    ),
+  };
+  const user: UserInfo = { name: 'username', roles: [] };
+
+  function docToChange(doc: DatabaseDocument): ChangeResult {
+    return {
+      doc,
+      id: doc._id,
+      changes: [{ rev: `rev-${doc._id}` }],
+      seq: `seq-${doc._id}`,
+    };
+  }
+
+  beforeEach(async () => {
+    mockCouchdbService = { get: () => undefined } as any;
+    mockRulesService = { getRulesForUser: () => undefined } as any;
+    const module: TestingModule = await Test.createTestingModule({
+      controllers: [ChangesController],
+      providers: [
+        ...authGuardMockProviders,
+        { provide: CouchdbService, useValue: mockCouchdbService },
+        { provide: RulesService, useValue: mockRulesService },
+        PermissionService,
+      ],
+    }).compile();
+
+    controller = module.get<ChangesController>(ChangesController);
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  it('should use the rules of the requesting user', () => {
+    getChangesResults([]);
+
+    expect(mockRulesService.getRulesForUser).toHaveBeenCalledWith(user);
+  });
+
+  it('should forward params', () => {
+    const params = { since: 'now', feed: 'continuous', limit: '500' };
+    getChangesResults([], params);
+
+    expect(mockCouchdbService.get).toHaveBeenCalledWith('some-db', '_changes', {
+      ...params,
+      include_docs: true,
+    });
+  });
+
+  it('should return all changes if user is allowed to read everything', async () => {
+    const res = await getChangesResults([{ subject: 'all', action: 'manage' }]);
+
+    expect(res.results.map((r) => r.id)).toEqual([
+      schoolDoc._id,
+      privateSchoolDoc._id,
+      childDoc._id,
+      deletedChildDoc._id,
+    ]);
+  });
+
+  it('should filter out changes for which the user does not have access', async () => {
+    const res = await getChangesResults([
+      { subject: 'School', action: 'read', conditions: { private: true } },
+      { subject: 'Child', action: 'manage' },
+    ]);
+
+    expect(res.results.map((r) => r.id)).toEqual([
+      privateSchoolDoc._id,
+      childDoc._id,
+      deletedChildDoc._id,
+    ]);
+  });
+
+  it('should always return deleted docs', async () => {
+    const res = await getChangesResults([
+      { subject: 'School', action: 'read' },
+    ]);
+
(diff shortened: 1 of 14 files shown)


src/restricted-endpoints/replication/changes/changes.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+      .pipe(map((res) => this.filterChanges(res, user)));
+  }
+
+  private filterChanges(
+    changes: ChangesResponse,
+    user: UserInfo,
+  ): ChangesResponse {
+    const ability = this.permissionService.getAbilityFor(user);
+    changes.results = changes.results
+      .filter((change) => this.canReadDoc(change.doc, ability))
```
TheSlimvReal, line 53: At the moment the initial `sync` call (not live sync) in the frontend finishes the first time an empty changes feed is returned.
This does not really create a problem as the remaining docs are synced through the live sync later.
We could however further reduce the amount of requests done by the frontend by already requesting and returning further changes here in the backend whenever not enough results remain in the returned changes.

Some questions remain, e.g. whether we always want to remain within the limit or whether we just ensure that at least limit docs are returned (but possibly more) to be more performant.

src/restricted-endpoints/replication/changes/changes.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+      .map((change) => omit(change, 'doc'));
```
TheSlimvReal, line 54: We could now enable the `include_docs: true` option again by not omitting the `doc` property in that case. But for the normal usage this does not add much benefit.

src/restricted-endpoints/replication/changes/changes.controller.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  it('should not return the document content', async () => {
+    const res = await getChangesResults([
+      { subject: 'School', action: 'read' },
+    ]);
+
+    res.results.forEach((r) => expect(r.doc).toBeUndefined());
+  });
+
+  function getChangesResults(rules: DocumentRule[], params?) {
```
sleidig, line 123: the tests confused me a bit initially. Maybe renaming this helper to something like `mockForRulesAndGetChanges` (or something simpler but more expressive than "getChangesResults" could help

sleidig: or maybe if `jest.spyOn(mockCouchdbService, 'get').mockReturnValue(of({ ...changes }));` can go into beforeEach and the other two lines are inlined in each test, the tests will be more immediately understandable?

src/restricted-endpoints/replication/changes/changes.controller.ts
sleidig: I think it would be good to do this not only for performance but also for cases that expect this to closely follow official specifications, e.g. make sure this also works without live sync.

I guess doing a "simple" recursive call in `changes()` to loop until we reach the end of the overall changes or the limit and cut of the returned result exactly at the limit wouldn't be extremely difficult?

src/restricted-endpoints/replication/changes/changes.controller.ts
sleidig: hmm, I guess permissions are ensured now either way :-) 
Shall we just leave it to the client then to request also with docs, if it wants?

sleidig: _deleted docs can not be requested with full body using _include_docs, can they?

sleidig (changes requested):
I love the simplicity of how you added this fairly powerful improvement :smiley: 

Looks good to me, although I would suggest we add code to avoid inadvertently stopping a sync early.


Will be interesting to observe how / whether this affects our server memory usage, etc.

src/restricted-endpoints/replication/changes/changes.controller.ts
TheSlimvReal: Depends a little. The way we normally do it, the document content is removed. But it is also possible to only delete a docs by setting `_deleted: true` manually and saving it. In this case the document content is kept and it would be returned here. 
The problem is, it is not possible to distinguish these cases. As we are only working with the former way I think this is a ok limitation.

src/restricted-endpoints/replication/changes/changes.controller.ts
TheSlimvReal: Not that simple but I see the point of following the specs a bit closer.

src/restricted-endpoints/replication/changes/changes.controller.ts
sleidig: if the doc content might be included, we shouldn't permit include_docs on changes. But that's alright I think.

TheSlimvReal: Ok so you suggest to always remove the `doc` property?

src/restricted-endpoints/replication/changes/changes.controller.ts
```diff
@@ -31,31 +33,68 @@ export class ChangesController {
...
-    user: UserInfo,
+    ability: DocumentAbility,
   ): ChangesResponse {
-    const ability = this.permissionService.getAbilityFor(user);
-    changes.results = changes.results
-      .filter((change) => this.canReadDoc(change.doc, ability))
-      .map((change) => omit(change, 'doc'));
+    changes.results = changes.results.filter(
+      ({ doc }) => doc._deleted || ability.can('read', doc),
+    );
```
sleidig, line 97:
```suggestion
    changes.results = changes.results.filter(
      ({ doc }) => doc._deleted || ability.can('read', doc),
    ).map((c) => ability.can('read', c.doc) ? c : omit(c, 'doc'));
```

src/restricted-endpoints/replication/changes/changes.controller.ts
```diff
@@ -31,31 +33,68 @@ export class ChangesController {
...
-        ...params,
-        include_docs: true,
-      })
-      .pipe(map((res) => this.filterChanges(res, user)));
+    @Query() params?: ChangesParams,
+  ): Promise<ChangesResponse> {
+    const ability = this.permissionService.getAbilityFor(user);
+    const change = { results: [] } as ChangesResponse;
+    let since = params?.since;
+    while (true) {
```
sleidig, line 44: might have been easier to read without the infinite loop, e.g. something like `while(!(!params?.limit || change.pending === 0 || change.results.length >= params.limit))` ?

But it's alright as it is.

src/restricted-endpoints/replication/changes/changes.controller.spec.ts
```diff
@@ -112,18 +111,85 @@ describe('ChangesController', () => {
...
+      .mockReturnValueOnce(
+        of(createChanges([schoolDoc, childDoc, childDoc], 2)),
+      )
+      .mockReturnValueOnce(of(createChanges([schoolDoc, childDoc, childDoc])));
+
+    const res = await controller.changes('some-db', user, { limit: 3 });
+
+    expect(res.pending).toBe(1);
+    expect(res.last_seq).toBe(docToChange(childDoc).seq);
+    res.results.forEach(({ id }) => expect(id).toBe(childDoc._id));
```
sleidig, line 160: I don't fully understand this test case: if the request is `limit: 3`, why is it getting something with "pending" then?
Maybe I am confused with the two mocked responses and reused items. Does this mean the total changes (global) are [schoolDoc, childDoc, childDoc, schoolDoc, childDoc, childDoc], i.e. length 6?

sleidig (approved):
Thanks for extending this.
Nice tests also :-)

src/restricted-endpoints/replication/changes/changes.controller.spec.ts
TheSlimvReal: Yes the global changes are correct. So the user requests 3 changes and can only read children docs and therefore receives`[childDoc, childDoc, childDoc]`. In the database there is one more change after the last `childDoc` that will not be returned so there is still one pending change.

TheSlimvReal: But your right, the `pending 2` was incorrect in the first changes feed.

src/restricted-endpoints/replication/changes/changes.controller.ts
TheSlimvReal: The problem is that I would need some kind of `repeat ... until` loop that at least runs once. 
That is a bit difficult to achieve with the while loop as it would require a additional variable which would be modified just where the current break is used. So I did not see much benefit.

src/restricted-endpoints/replication/changes/changes.controller.ts
TheSlimvReal: Good idea, I adjusted it a little bit to not have a inconsistent structure at the end (some changes have a `doc` property while others don't)

sleidig (approved):

PR #114, opened 2023-08-23
Delete endpoints

TheSlimvReal:
With the recent change of the proxy setup towards explicit endpoints I have forgot to implement the delete endpoints. They were actually not secured and could have been used to delete docs without necessary permissions.
This has been changed now.


src/restricted-endpoints/attachment/attachment/attachment.controller.ts
```diff
@@ -110,9 +115,29 @@ export class AttachmentController {
     this.proxy(request, response, () => undefined);
   }
 
+  /**
+   * Deletes an attachment if the user has `read` permissions.
```
sleidig, line 119:
```suggestion
   * Deletes an attachment if the user has `delete` permissions.
```
?

src/restricted-endpoints/document/document.controller.ts
```diff
@@ -65,4 +67,22 @@ export class DocumentController {
...
+   * @param queryParams additional params that will be forwarded
+   */
+  @Delete()
+  async deleteDocument(
+    @Param('db') db: string,
+    @Param('docId') docId: string,
+    @User() user: UserInfo,
+    @Query() queryParams?: QueryParams,
+  ) {
+    return this.documentService.deleteDocument(db, docId, user, queryParams);
```
sleidig, line 86: Slightly "inconsistent" between the controllers, as the attachment.controller enforces permissions (I like the simplicity of that) and the document.controller leaves it to the service to enforce permissions. Is there a technical reason for that?

Not a major issue though, just making the remark here.

sleidig (approved):
looks nice, good job!

src/restricted-endpoints/document/document.controller.ts
TheSlimvReal: Good point, the document controller is not really adding anything so I merged the service in here

PR #118, opened 2023-11-07
Backend fixes

TheSlimvReal:
On the dev server there is currently an issue where the changes are requested continuously. 
This is because the `last_seq` is not returned correctly from the backend due to a special permissions-data setup.

Also we sometimes see error messages on the backend where the rules object fetch times out (`AxiosError: maxContentLength size of -1 exceeded`.
Normally this is a longpoll and it should be possible to have a open connection for multiple minutes but the axios request times out after one minute. The same happens when using native NodeJS `fetch`. If the same request is send from the command line using curl, everything works fine (I had a exception after around 9 minutes). I did not find a way to change anything the app. I don't think the servers nginx setup is the issue because the requests work fine without the backend.
I now reduced the timeout for CouchDB to 50s so it does not throw errors anymore.

### DIFF
--- a/src/permissions/rules/rules.service.ts
+++ b/src/permissions/rules/rules.service.ts
@@ -41,14 +41,24 @@ export class RulesService {
         since: this.lastSeq,
         include_docs: true,
         doc_ids: JSON.stringify([Permission.DOC_ID]),
+        // requests somehow time out after 1 minute
+        timeout: 50000,
       }),
     );
+    const before = new Date().getTime();
 
     return getParams
       .pipe(
-        concatMap((params) =>
-          this.couchdbService.get<ChangesResponse>(db, '_changes', params),
-        ),
+        concatMap((params) => {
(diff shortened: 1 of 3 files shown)


src/permissions/rules/rules.service.ts
```diff
@@ -41,14 +41,24 @@ export class RulesService {
...
 
     return getParams
       .pipe(
-        concatMap((params) =>
-          this.couchdbService.get<ChangesResponse>(db, '_changes', params),
-        ),
+        concatMap((params) => {
+          const date = new Date().getTime();
+          const diff = (date - before) / 1000;
+          console.log(diff.toFixed(2), 'restarting');
```
sleidig, line 55: remove console.log before merge? Or is this useful debugging info for production systems also?

sleidig (approved):
Looks good to me.

Thanks a lot for troubleshooting that and figuring out these fixes!

PR #127, opened 2024-02-20
feat: add support for _find and head requests

tomwwinter:


src/restricted-endpoints/document/document.controller.spec.ts
```diff
@@ -12,11 +12,18 @@ import { CouchdbService } from '../../couchdb/couchdb.service';
...
 import { UnauthorizedException } from '@nestjs/common';
+import { Request as Req } from 'express';
 
 describe('DocumentController', () => {
   let controller: DocumentController;
   let mockPermissionService: PermissionService;
   let mockCouchDBService: CouchdbService;
+
+  const headers = {};
+  const mockReq = {
```
sleidig, line 23: can we add a test case checking for the new functionality also, please? replication-backend is a production system and key service for us :blush:

src/restricted-endpoints/document/document.controller.ts
```diff
@@ -40,6 +43,37 @@ export class DocumentController {
...
+  ): Promise<DatabaseDocument> {
+    const userAbility = this.permissionService.getAbilityFor(user);
+    const document = await firstValueFrom(
+      this.couchdbService.get(db, docId, queryParams),
+    );
+    if (userAbility.can('read', document)) {
+      if (document._rev) {
+        req.res.setHeader('etag', document._rev);
+      }
+      return document;
```
sleidig, line 71: according to specs this shouldn't return a document, only the headers, right?

Of course, in order to check permissions we do need to load the doc anyway ... But I would tend to stick to our approach and implement the actual CouchDB API?

src/restricted-endpoints/replication/bulk-document/bulk-document.service.ts
```diff
@@ -106,6 +107,38 @@ export class BulkDocumentService {
...
+    const response = await firstValueFrom(
+      this.couchdbService.post<AllDocsResponse>(
+        db,
+        '_all_docs',
+        allDocsRequest,
+        {
+          include_docs: true,
+        },
+      ),
+    );
```
sleidig, line 128: why do we need an additional all_docs request here? Can't we just check permissions on the response of the "_find"?

src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
```diff
@@ -65,6 +66,31 @@ export class BulkDocEndpointsController {
...
+   * @param db name of the database to which the documents should be uploaded
+   * @param body search query object
+   * @param user logged in user
+   * @returns BulkDocsResponse list of documents matching search query
+   */
+  @Post('/:db/_find')
+  @ApiOperation({
+    description: `Find multiple documents with a single request.`,
+  })
+  find(
```
sleidig, line 82: Let's write some simple unit tests for the find controller and/or service

sleidig (changes requested):

src/restricted-endpoints/document/document.controller.ts
```diff
@@ -44,34 +45,42 @@ export class DocumentController {
...
-        req.res.setHeader('etag', document._rev);
-      }
-      return document;
-    } else {
+
+    if (
+      !userAbility.can('read', {
+        _id: docId,
+      })
+    ) {
```
sleidig, line 70: @tomwwinter If we have a permission rule like "can read all docs with location=Berlin", won't this return 401 even for allowed docs?

Where is the HEAD request used? How critical is a possible difference between the head and actual get?

tomwwinter: Good point, the field permissions are a litte bit tricky... lets discuss this today

PR #140, opened 2024-04-11
feat: remove client errors from sentry logging

tomwwinter:


src/sentry.configuration.ts
```diff
@@ -63,8 +60,11 @@ function configureLoggingSentry(
 
     beforeSend: (event, hint) => {
       const error = hint.originalException;
-      if (error instanceof UnauthorizedException) {
-        console.log(error);
+      if (
```
sleidig, line 63: we generally hardly have any logs in the container itself. We could "abuse" this function to log all errors to console as a side effect ...

sleidig (approved):

src/sentry.configuration.ts
tomwwinter: I'd rather add specific logs then, otherwise we'll just spam up our hard disk with (mostly) useless information.

PR #141, opened 2024-04-16
feat: semantic versioning

tomwwinter:


.dockerignore
```diff
@@ -0,0 +1,5 @@
+Dockerfile
+.dockerignore
```
sleidig, line 2: why do we ignore the Dockerfile?

.github/workflows/pull-request-update.yml
```diff
@@ -1,28 +0,0 @@
-name: Pipeline
-on: pull_request
```
sleidig, line 2: should we still test to build on PRs?

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,36 @@
...
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Login to GitHub Container Registry
```
sleidig, line 23: If we switch to GitHub Container Registry from DockerHub, let's check all docs (espcially ndb-setup repo) to update links.

sleidig (approved):

.dockerignore
tomwwinter: It just means, that we will not copy the Dockerfile into the Container.

.github/workflows/pull-request-update.yml
tomwwinter: The test just moved to another workflow, we still run the test on PRs

.github/workflows/build-release.yml
tomwwinter: Yes, i will update all the docs.

PR #147, opened 2024-05-21
feat: support for user linked project permissions

tomwwinter:
related: https://github.com/Aam-Digital/internal/issues/21

### DIFF
--- a/src/auth/guards/jwt-bearer/jwt-bearer.strategy.ts
+++ b/src/auth/guards/jwt-bearer/jwt-bearer.strategy.ts
@@ -4,6 +4,8 @@ import { Injectable } from '@nestjs/common';
 import { UserInfo } from '../../../restricted-endpoints/session/user-auth.dto';
 import { ConfigService } from '@nestjs/config';
 import { AuthModule } from '../../auth.module';
+import { CouchdbService } from '../../../couchdb/couchdb.service';
+import { firstValueFrom } from 'rxjs';
 
 /**
  * Authenticate a user with a foreign bearer JWT using the {@link AuthModule.JWT_PUBLIC_KEY}.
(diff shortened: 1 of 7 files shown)


src/auth/guards/jwt-bearer/jwt-bearer.strategy.ts
```diff
@@ -13,14 +15,25 @@ export class JwtBearerStrategy extends PassportStrategy(
...
-    return new UserInfo(data.username, data['_couchdb.roles']);
+    const user = await firstValueFrom(
+      this.couchdbService.get('app', data['username']),
+    ).catch(() => {});
+
+    return new UserInfo(
+      data.username,
+      data['_couchdb.roles'],
+      user && user['projects'] ? user['projects'] : [],
+    );
```
sleidig, line 37: small code block, but also for future proofing I would suggest we put the UserInfo creation in a service rather than duplicate it.

sleidig (changes requested):
Looks good so far, but isn't this currently still missing the core of the logic, i.e. that the project field is used to filter out docs during replication?
Or can "projects" now somehow be used in configured rules?

sleidig:
> Looks good so far, but isn't this currently still missing the core of the logic, i.e. that the project field is used to filter out docs during replication? Or can "projects" now somehow be used in configured rules?

Okay, understood now that this simply works out of the box :smile: 
(placeholder is available as `${user.projects}` for the permission config)

- [ ] Let's add the details about this placeholder to the Dev Docs: https://aam-digital.github.io/ndb-core/documentation/additional-documentation/concepts/user-roles-and-permissions.html
- [ ] Let's remove the documentation about permissions config from the README here, replacing with a link to the docs

sleidig (approved):
Looks good!

Could still remove that code duplication here https://github.com/Aam-Digital/replication-backend/pull/147#discussion_r1631002196
but we can also do this once we touch that code again.

PR #159, opened 2024-09-10
fix: check all doc access for app-attachment against the app entity

sleidig:
fixes https://github.com/Aam-Digital/ndb-core/issues/2558
fixes https://github.com/Aam-Digital/ndb-core/issues/1985


tomwwinter:
Functionality looks good, but we still have some bugs.

- Application is re-syncing after reload
- After reloading on a child entity, i got this error: 
image

tomwwinter (approved):

PR #171, opened 2025-02-11
Support user.id placeholder for permissions

sleidig:
see https://github.com/Aam-Digital/ndb-core/issues/2847


src/auth/guards/jwt-cookie/jwt-cookie-strategy.service.ts
```diff
@@ -32,7 +32,9 @@ export class JwtCookieStrategy extends PassportStrategy(
       this.couchdbService.get('app', data['username']),
     ).catch(() => {});
 
+    // TODO
```
tomwwinter, line 35: What is missing here?

sleidig:
```suggestion
```

:see_no_evil:   I just forgot to remove the TODO comment that I had added after screening the code

tomwwinter (approved):

PR #240, opened 2026-04-30
Fix/permission init issues

### LINKED ISSUE #238: Harden RulesService to fail-closed when no valid permission config is available
## Problem

`RulesService` can leave `this.permission === undefined` after startup. When that happens, [`getRulesForUser`](https://github.com/Aam-Digital/replication-backend/blob/master/src/permissions/rules/rules.service.ts) returns

```ts
return [{ subject: 'all', action: 'manage' }];
```

as a fallback, i.e. **every authenticated user is granted full read/write on every document**.

Paths that can leave `this.permission` undefined today:

1. `loadInitialPermissions` succeeds but the permission doc is missing/invalid (404, empty `data`, malformed payload). The `catch` only re‑throws on 401/403; other errors are logged at WARN and the service starts fail‑open.
2. CouchDB is reachable but slow/transient errors during initial fetch — same path as above.
3. `app.listen()` happens after `onModuleInit` resolves, but in the failure cases above it resolves *successfully* with `permission` still undefined → requests can be served from the fail‑open fallback.

The recently introduced exponential backoff on the changes feed (#235) widens the window during which permissions cannot recover via the live feed.

## Proposed hardening

Two‑layer fix in `RulesService`:

1. **Block startup until a valid permission config is loaded.** Retry transient errors (network, 5xx, 404, empty/invalid payload) with bounded exponential backoff; throw on persistent failure or auth errors. The container will then crash loudly on misconfiguration rather than silently fail‑open, and Nest will not call `app.listen()` until permissions are present.
2. **Replace the fail‑open fallback in `getRulesForUser`** (`return [{ subject: 'all', action: 'manage' }]`) with deny‑all (`return []`), as defense in depth. Consistent with the anonymous branch which already returns `[]` when no `public` rules are defined.

Runtime behaviour on `Config:Permissions` deletion (tombstone) is unchanged: the previous in‑memory copy is kept (already implemented in `watchPermissionChanges`).

## Out of scope

- Reserved role names (`default`, `public`) collision — tracked separately in #236.
- Wider permission/endpoint refactors.

sleidig:
- closes #238

## Summary by CodeRabbit

* **Documentation**
  * Added detailed docs describing the permission service startup flow, failure outcomes, retry behavior, and fail-closed safety.

* **New Features**
  * Added a reusable exponential-backoff utility for retry handling.

* **Bug Fixes**
  * Permission loading now retries with backoff, boots into an admin-only bootstrap mode when missing, and denies access until rules load.
  * Improved feed error logging and retry behavior.

* **Tests**
  * Expanded tests for backoff, feed resilience, and permission startup edge cases.

* **Chores**
  * Build script now removes TypeScript incremental artifacts; TypeScript target updated.


tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es2017",
+    "target": "es2022",
```
Abhinegi2, line 9: Any specific reason for migrating to ES2022?

sleidig: yes, using the simpler `Object.hasOwn` is only available in newer node versions. As we run within our own docker image that already has a recent LTS version, this will not have much risk

Abhinegi2 (approved):

src/permissions/rules/rules.service.ts
```diff
@@ -47,32 +67,103 @@ export class RulesService implements OnModuleInit {
...
+          );
+        }
+
+        // Do not overwrite permissions that may have arrived from the live feed already.
+        if (this.permission === undefined) {
+          this.permission = data;
+        }
+        return;
+      } catch (error) {
+        if (RulesService.isAuthFailure(error)) {
```
sleidig: should this really treat auth failures specially without a few retries?!

PR #257, opened 2026-06-02
feat(audit): record entity change log to a separate audit database

### LINKED ISSUE #4026: Change Log: record changes in backend
- implementation phase 1 of #490

Abhinegi2:
## Summary

Adds backend recording of every entity change (audit / changelog) to the replication-backend, per the implementation plan on the issue. Every authenticated, permitted, successful entity write is recorded to a separate, client-inaccessible `<db>-audit` database, so that **what changed, by whom and when** can be reconstructed for audit/legal purposes.

Gated behind the `AUDIT_ENABLED` flag (**default off**) — a complete no-op when disabled, with no behavior change on any write path.

## What it does

- **Records** create / update / delete on both write paths: bulk replication (`_bulk_docs`) and single `PUT`/`DELETE`. The component that performs the write also triggers the audit ("the writer audits"), reusing the before-state already fetched for permission checks.
- **Audit record** carries: `entityId`, source `database`, `operation`, new `rev` + `parentRev`, a **server-set** `timestamp`, the **authenticated** `user` (`{id, name, roles}`), and a [`jsondiffpatch`](https://www.npmjs.com/package/jsondiffpatch) `diff`. The client's local `_updatedAt`/`_updatedBy` are kept inside the diff; only internal CouchDB noise (`_rev`, `_revisions`, `_conflicts`, `_attachments`) is excluded.
- **Seamless activation:** the first change to a previously-unaudited entity additionally writes a full-snapshot `baseline` record, so history is not lost when the feature is switched on for an existing system. No migration needed.
- **Protection:** a global guard (`APP_GUARD`) rejects any client request to a `<db>-audit` database (every route), so audit data is read-only to users via the proxy.

## Audit DB naming

Derived, hard-wired: writes to `app` are recorded in `app-audit`. Each audit db is auto-created on first write.

## Known limitation (documented in README)

Only revisions *pushed* by clients are captured. PouchDB sends only **leaf** revisions with their ancestry, so intermediate same-device/offline edits never reach the backend and cannot be audited by any backend component. The guarantee is conflict-branch capture plus a trustworthy authenticated author and server timestamp.

## Testing

- Unit tests for: record building + diff (create/update/delete), kept/excluded fields, baseline-on-first-change (and not on second), skipping non-replicable / conflicted / audit-db docs, server timestamp + authenticated user, `new_edits:false`/`true` success detection, and the access guard.
- Verified end-to-end against a real local stack (`aam-digital-developer-stack`): edited a record in the app as a Keycloak user → got one `baseline` + one `update` record with the correct authenticated user, server timestamp, `parentRev`, and a diff matching the edit; client access to `*-audit` returns `403`.

## Deployment

`AUDIT_ENABLED` env var (default off) on the replication-backend service. The `<db>-audit` databases are auto-created. ndb-setup wiring is intentionally left out of this PR.

Closes Aam-Digital/ndb-core#4026

## Summary by CodeRabbit

* **New Features**
  * Optional audit logging for entity writes with per-change records, automatic baseline snapshots on first audit, and per-record user/timestamp/revision/diff metadata (disabled by default).

* **Documentation**
  * Added "Audit / Changelog Recording" section documenting configuration, behavior, and known limitations.

* **Chores**
  * Added diffing dependency and security-ignore entry for a known advisory.

* **Tests**
  * Expanded coverage for audit, baseline, and bulk-write scenarios.

* **Refactor**
  * Document write and bulk-write flows delegated to centralized services to standardize auditing.


sleidig:
functionally looks good at first glance!

Did you finalize the code already, @Abhinegi2 ? Then I will do a review and slightly more thorough testing

Abhinegi2:
> functionally looks good at first glance!
> 
> Did you finalize the code already, @Abhinegi2 ? Then I will do a review and slightly more thorough testing

Yes you can also start code review

src/audit/audit-record.dto.ts
```diff
@@ -0,0 +1,90 @@
...
+  operation: AuditOperation;
+}
+
+/**
+ * The persisted audit document, stored in the derived `<db>-audit` database.
+ *
+ * `_id` encodes the entity id and timestamp for performant per-entity range
+ * queries: `ChangeAudit:<entityId>:<ISO-timestamp>:<new-rev>`.
+ */
+export interface AuditRecord {
```
sleidig: Should we maybe call this `ChangeAuditEntity` to intuitively signal that this is the one saved in the DB and align with the prefix used for ID and permissions?

Or actually I would maybe prefer `AuditRecordEntity` and change the prefix accordingly?

src/audit/audit-record.dto.ts
```diff
@@ -0,0 +1,90 @@
...
+  parentRev?: string;
+  /** SERVER-set time (never trusted from the client body) */
+  timestamp: string;
+  /** SERVER-set from the authenticated UserInfo (never trusted from the body) */
+  user: { id: string; name: string; roles: string[] };
+  /**
+   * For create/update/delete: a jsondiffpatch delta (winning-rev -> written-rev).
+   * For baseline: the full previous document (there is nothing to diff against).
+   */
+  diff?: any;
```
sleidig, line 86:
```suggestion
  /** the changed doc's `_id`, e.g. `Child:123` */
  entityId: string;

  /** source db name, e.g. `app` */
  database: string;

  operation: AuditOperation;

  /** new `_rev` of the written revision */
  rev?: string;

  /** the written rev's parent, from the submitted doc's `_revisions` ancestry */
  parentRev?: string;

  /** SERVER-set time (never trusted from the client body) */
  timestamp: string;

  /** SERVER-set from the authenticated UserInfo (never trusted from the body) */
  user: { id: string; name: string; roles: string[] };

  /**
   * For create/update/delete: a jsondiffpatch delta (winning-rev -> written-rev).
   * For baseline: the full previous document (there is nothing to diff against).
   */
  diff?: any;
```

src/audit/audit.config.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * CASL subject prefix for audit record `_id`s (e.g.
+   * `ChangeAudit:Child:123:<ts>:<rev>`). The proxy derives the permission
+   * subject from the `_id` prefix (`detectDocumentType` = `_id.split(':')[0]`),
+   * so this dedicated subject lets a single rule govern audit records — keeping
+   * them read-only and un-forgeable regardless of the source entity's
+   * permissions, and reachable as an ordinary read-only remote DB by the
+   * history-viewing UI (see #4027).
+   */
+  static readonly AUDIT_SUBJECT = 'ChangeAudit';
```
sleidig: aligning with the code structure of ndb-core, maybe make this a static property of a `ChangeAuditEntity` class?

src/audit/audit.service.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  );
+  // _id prefix makes the proxy classify this as subject `ChangeAudit`
+  expect(records[0]._id.split(':')[0]).toBe('ChangeAudit');
+  expect(records[0].diff).toBeDefined();
+  expect(couchdb.post).toHaveBeenCalledWith('app-audit', '_bulk_docs', {
+    docs: records,
+  });
+});
+
+it('keeps _updatedAt/_updatedBy and excludes _rev/_revisions in the diff', async () => {
```
sleidig: are these really _updatedAt? If you talk about the Entity objects created in the frontend, the properties should be updated and created?

Or does CouchDB really store some hidden _updatedAt property?

src/audit/audit.service.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+  await service.record('app', [entry], user);
+
+  const records = postedRecords(couchdb);
+  expect(records).toHaveLength(2);
+  expect(records[0].operation).toBe('baseline');
+  expect(records[0].diff).toEqual({ _id: 'Child:1', name: 'A' });
+  expect(records[1].operation).toBe('update');
+});
+
+it('does not write a baseline when the entity already has an audit record', async () => {
```
sleidig, line 143: can we also assert some test case(s) that a correct diff is created and saved for updates?

src/audit/audit.service.ts
```diff
@@ -0,0 +1,261 @@
...
+const IGNORED_DIFF_FIELDS = [
+  '_rev',
+  '_revisions',
+  '_conflicts',
+  '_attachments',
+];
+
+/**
+ * Records every entity write to a separate, client-inaccessible `<db>-audit`
+ * database (see issue #4026). Gated behind the {@link AuditConfig.AUDIT_ENABLED_ENV}
```
sleidig:
```suggestion
 * database. Gated behind the {@link AuditConfig.AUDIT_ENABLED_ENV}
```

Let's not reference issues unless it is an open TODO

src/audit/audit.service.ts
```diff
@@ -0,0 +1,261 @@
...
+ * flag; a complete no-op when disabled.
+ *
+ * Writes are best-effort-but-logged for v1: a failed audit write never blocks
```
sleidig:
```suggestion
 * Writes are best-effort-but-logged: a failed audit write never blocks
```

src/audit/audit.service.ts
```diff
@@ -0,0 +1,261 @@
...
+      // SERVER-set timestamp, never trusted from the client body
+      const timestamp = new Date().toISOString();
+      const records: AuditRecord[] = [];
+
+      for (const entry of relevant) {
+        const entityId = entry.newDoc._id;
+        if (!entityId) {
+          // guaranteed by the isReplicableId filter above; narrows the type
+          continue;
+        }
```
sleidig: so is this a realistic path? Or something that should cause a warning (or that actually never happens)?

src/audit/audit.service.ts
```diff
@@ -0,0 +1,261 @@
...
+        records.push(this.buildRecord(db, entityId, entry, user, timestamp));
+      }
+
+      await firstValueFrom(
+        this.couchdbService.post(auditDb, '_bulk_docs', { docs: records }),
+      );
+    } catch (err) {
+      // best-effort: log but never fail the original write
+      this.logger.error(
+        `Failed to write audit records for db '${db}': ${err}`,
```
sleidig: if err can be an object, wouldn't this show `[Object object]` in the logs?

src/restricted-endpoints/document/document.controller.ts
```diff
@@ -40,8 +36,9 @@ import { AxiosResponse } from 'axios';
 @Controller('/:db/:docId')
 export class DocumentController {
   constructor(
-    private couchdbService: CouchdbService,
-    private permissionService: PermissionService,
+    private readonly couchdbService: CouchdbService,
+    private readonly permissionService: PermissionService,
```
sleidig, line 40: does this still need couchdB and permissions directly here also?

src/restricted-endpoints/replication/bulk-document/bulk-document.service.spec.ts
```diff
@@ -276,6 +280,82 @@ describe('BulkDocumentService', () => {
...
+    await service.handleBulkDocs(request, normalUser, 'app');
+
+    expect(mockAuditService.record).toHaveBeenCalledTimes(1);
+    const [db, entries, user] = mockAuditService.record.mock.calls[0];
+    expect(db).toBe('app');
+    expect(user).toBe(normalUser);
+    expect(entries).toHaveLength(1);
+    expect(entries[0].operation).toBe('update');
+    expect(entries[0].newRev).toBe('2-new');
+    expect(entries[0].existingDoc._id).toBe('Child:1');
```
sleidig: are we also assuring the actual write (to app DB) is happening here?

src/restricted-endpoints/replication/bulk-document/bulk-document.service.spec.ts
```diff
@@ -276,6 +280,82 @@ describe('BulkDocumentService', () => {
...
+  });
+
+  it('audits with rev from the response on the new_edits:true path', async () => {
```
sleidig: I don't fully understand what that means 😊 

Could you explain?

src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
sleidig: can we rename this to `bulkDocumentService` for clarity?

src/restricted-endpoints/replication/bulk-document/bulk-document.service.ts
```diff
@@ -97,29 +139,110 @@ export class BulkDocumentService {
...
+        errorIds.add(result.id);
+      } else {
+        revById.set(result.id, (result as DocSuccess).rev);
+      }
+    }
+    return { errorIds, revById };
+  }
+
+  /** Build a single audit entry, or null if the doc was not successfully written. */
+  private toAuditEntry(
```
sleidig: couldn't/shouldn't this move to the AuditService?

README.md
```diff
@@ -86,6 +86,86 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
...
+- Any **client** write that targets a `ChangeAudit:` document (to the audit DB
+  _or_ a forged one in the main app DB) is dropped, because no rule grants
+  `create`/`update`/`delete` on `ChangeAudit`. Rules are global (keyed on the
+  subject, not the DB name), so this holds across `_bulk_docs` / `_all_docs` /
+  `_changes` / `_bulk_get` / `_find`.
+- A client **read** of audit records is allowed only where a
+  `{ subject: "ChangeAudit", action: "read" }` rule is granted (the proxy filters
+  reads via `ability.can('read', doc)`). This is what lets the history-viewing UI
+  (see [#4027](https://github.com/Aam-Digital/ndb-core/issues/4027)) read the
+  audit DB as an ordinary read-only remote database.
```
sleidig:
```suggestion
  read the audit DB as an ordinary read-only remote database.
```

README.md
```diff
@@ -86,6 +86,86 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
...
+> the read rule only to privileged roles who already see full records.
+> Field-level redaction would require a bespoke read endpoint instead.
+
+### Known limitation
+
+Only revisions _pushed_ by clients are captured. PouchDB sends only **leaf**
+revisions with their ancestry, so intermediate same-device/offline edits never
+reach the backend and cannot be audited by any backend component. The genuine
+guarantee is conflict-branch capture and a trustworthy authenticated author and
+server timestamp — not a complete per-keystroke history.
```
sleidig:
```suggestion
server timestamp.
```

per keystroke is anyway not saved

README.md
```diff
@@ -86,6 +86,86 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
...
+
+The CouchDB admin reverse proxy at `/couchdb/` (admin-only) bypasses the
+permission engine; audit protection relies on CouchDB only being reachable via
+the proxy on the internal network.
+
```
sleidig: This is true for everything, no need to mention here, I think.
```suggestion
```

sleidig (commented):
Looks really good overall!

src/restricted-endpoints/document/document.controller.ts
Abhinegi2: Yes, both are still used.

src/audit/audit.service.ts
Abhinegi2: It never happens in practice — the `relevant` list above is already filtered by `isReplicableId(entry.newDoc._id)`, which guarantees `_id` is a truthy replicable string. The guard is purely a TypeScript narrowing of `string | undefined` → `string` for the downstream `buildAuditId`/`buildBaseline` calls, so it's not an error condition worth warning about. Happy to drop it for a non-null assertion

Abhinegi2: Dropped it — replaced the unreachable guard with a non-null assertion (`entry.newDoc._id!`) in ed86771.

src/audit/audit-record.dto.ts
```diff
@@ -0,0 +1,113 @@
...
+ * queries: `AuditRecord:<entityId>:<ISO-timestamp>:<new-rev>`.
+ *
+ * Owns the {@link AuditRecordEntity.TYPE} subject prefix (mirroring how ndb-core
+ * entities expose their type): the proxy derives the CASL subject from the `_id`
+ * prefix (`detectDocumentType` = `_id.split(':')[0]`), so audit docs are
+ * classified as the dedicated subject `AuditRecord`, not as the source entity.
+ */
+export class AuditRecordEntity {
+  /** CASL subject + `_id` prefix for all audit records */
+  static readonly TYPE = 'AuditRecord';
```
sleidig:
```suggestion
  static readonly ENTITY_TYPE = 'AuditRecord';
```

align completely with ndb-core class properties

src/audit/audit-record.dto.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /** the changed doc's `_id`, e.g. `Child:123` */
+  entityId!: string;
+
+  /** source db name, e.g. `app` */
+  database!: string;
+
+  operation!: AuditOperation;
+
+  /** new `_rev` of the written revision */
```
sleidig:
```suggestion
  /**
    * new `_rev` of the written revision.
    *  (not the _rev of this AuditRecordEntity in the DB)
    */
```

src/audit/audit-record.dto.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Owns the {@link AuditRecordEntity.TYPE} subject prefix (mirroring how ndb-core
+ * entities expose their type): the proxy derives the CASL subject from the `_id`
+ * prefix (`detectDocumentType` = `_id.split(':')[0]`), so audit docs are
+ * classified as the dedicated subject `AuditRecord`, not as the source entity.
+ */
+export class AuditRecordEntity {
+  /** CASL subject + `_id` prefix for all audit records */
+  static readonly TYPE = 'AuditRecord';
+
+  _id!: string;
```
sleidig, line 57:
```suggestion
  /**  the internal uuid of this AuditRecordEntity doc in the DB */
  _id!: string;
```

src/audit/audit.service.ts
```diff
@@ -0,0 +1,367 @@
...
+    if (!this.enabled) {
+      return;
+    }
+    if (isAuditDb(db)) {
+      // never audit the audit db itself
+      return;
+    }
+
+    const relevant = (entries ?? []).filter(
+      (entry) => entry?.newDoc && isReplicableId(entry.newDoc._id),
```
sleidig, line 121: why are we filtering for newDoc here? is this something different than the "is new entity" check (which would skip update/delete)?

src/audit/audit.service.ts
```diff
@@ -0,0 +1,367 @@
...
+    );
+    if (relevant.length === 0) {
+      return;
+    }
+
+    try {
+      const auditDb = auditDbFor(db);
+      await this.ensureDb(auditDb);
```
sleidig, line 129: maybe more economical to do this check only once during startup?

If we only start processing changes after such an init is completed, we could probably even remove any monitoring of ensuredDbs?

src/audit/audit.service.ts
```diff
@@ -0,0 +1,367 @@
...
+   * @param db source database name (e.g. `app`)
+   * @param entries the writes performed against `db`
+   * @param user the authenticated user (server-trusted "who")
+   */
+  async record(
+    db: string,
+    entries: AuditEntry[],
+    user: UserInfo,
+  ): Promise<void> {
+    if (!this.enabled) {
```
sleidig: I feel like this logically belongs more to the module than this service ... moving this check into the DI, injecting a NoopService if not enabled.

(already done and committed)

src/audit/audit.service.ts
```diff
@@ -0,0 +1,367 @@
...
+   * A baseline snapshot is needed when an existing entity is changed for the
+   * first time after the feature was activated (no prior audit record).
+   * Creates skip the baseline — the create record is itself the anchor.
+   */
+  private async needsBaseline(
+    auditDb: string,
+    entityId: string,
+    entry: AuditEntry,
+  ): Promise<boolean> {
+    if (!entry.existingDoc || entry.operation === 'create') {
```
sleidig, line 267: when is this ! existingDoc? Would that be the case for deletions? And if so, do we not need a baseline to be able to restore from delete?

src/restricted-endpoints/document/document-write.service.ts
```diff
@@ -0,0 +1,170 @@
...
+    });
+    if (permittedFields.length > 0) {
+      // Updating some properties
+      const updatedFields = pick(newDoc, permittedFields);
+      return Object.assign(oldDoc, updatedFields);
+    } else {
+      // Updating whole document
+      return newDoc;
+    }
+  }
```
sleidig, line 169: Is this really part of the current replication-backend functionaly yet? I remember we did some experiments about permissions that allow changes only to some fields of a doc, but put this on hold due to very difficult edge cases.

README.md
```diff
@@ -86,6 +86,81 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
 Additionally, a separate check on the client side is necessary that cleans up the local database whenever a client looses read permissions for a document.
 A example for how this could look can be found [here](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/permissions/permission-enforcer/permission-enforcer.service.ts).
 
+## Audit / Changelog Recording
+
+The backend can record a tamper-resistant change log of every entity write, so
+that "what changed, by whom and when" can be reconstructed for audit/legal
+purposes (see [issue #4026](https://github.com/Aam-Digital/ndb-core/issues/4026)).
```
sleidig: don't reference issues, except for open //TODO comments
```suggestion
purposes.
```

README.md
```diff
@@ -86,6 +86,81 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
...
+- On each successful write (bulk `_bulk_docs` replication, single `PUT`, and
+  `DELETE`), one record is written to a separate database derived from the
+  source db name: writes to `app` are recorded in `app-audit`. The convention
+  is hard-wired, and each `<db>-audit` database is auto-created on first write.
+- Each record's `_id` is `AuditRecord:<entityId>:<ISO-timestamp>:<rev>`. The
+  `AuditRecord:` prefix is load-bearing: the proxy derives the CASL subject from
+  the `_id` prefix (`detectDocumentType` = `_id.split(':')[0]`), so audit records
+  are classified as the dedicated subject `AuditRecord` — not as the source
+  entity (`Child`, ...). This is what keeps them un-forgeable and governed by a
+  single rule (see Protection).
```
sleidig: That's technical implementation, shouldn't be part of the high-level README, in my opinion.
```suggestion
```

README.md
```diff
@@ -86,6 +86,81 @@ A swagger / OpenAPI interface can be visited at `/api/` which shows all endpoint
...
+  permission-checked endpoints — so recording always succeeds.
+- Any **client** write that targets a `AuditRecord:` document (to the audit DB
+  _or_ a forged one in the main app DB) is dropped, because no rule grants
+  `create`/`update`/`delete` on `AuditRecord`. Rules are global (keyed on the
+  subject, not the DB name), so this holds across `_bulk_docs` / `_all_docs` /
+  `_changes` / `_bulk_get` / `_find`.
+- A client **read** of audit records is allowed only where a
+  `{ subject: "AuditRecord", action: "read" }` rule is granted (the proxy filters
+  reads via `ability.can('read', doc)`). This is what lets the history-viewing UI
+  read the audit DB as an ordinary read-only remote database.
```
sleidig:
```suggestion
The `<db>-audit` database is reachable through the
normal proxy and governed entirely by CASL on the `AuditRecord` subject:

- The system's own audit writes use the proxy's admin credentials and are
  written **directly** to CouchDB via `CouchdbService`, bypassing the
  permission-checked endpoints, so recording always succeeds.
- Any **client** write that targets a `AuditRecord:` document (to the audit DB
  _or_ a forged one in the main app DB) is dropped.
- A client **read** of audit records is allowed only where a
  `{ subject: "AuditRecord", action: "read" }` rule is granted (the proxy filters
  reads via `ability.can('read', doc)`). This is what lets the history-viewing UI
  read the audit DB as an ordinary read-only remote database.
```

sleidig (commented):
some final clarifications, should be good to go soon :-)

src/audit/audit.service.ts
sleidig: please do review and test this once, however :blush:

src/audit/audit.service.ts
Abhinegi2: It'\''s not the "is new entity" check — updates and deletes both still carry a `newDoc` (a delete'\''s `newDoc` is the tombstone with `_deleted: true`), so they'\''re not skipped here. The create-vs-update distinction is made later via `existingDoc` (in `needsBaseline`/`toAuditEntry`). This `entry?.newDoc` clause is just a defensive guard against malformed/empty entries plus type-narrowing so `entry.newDoc._id` is reachable for `isReplicableId`. Added a JSDoc comment to make this clear.

src/audit/audit.service.ts
Abhinegi2: It'\''s already effectively once-per-db: `ensureDb` short-circuits on the in-memory `ensuredDbs` set, so only the first write to a given db calls CouchDB (`createDb`) — every later write is a free `Set.has()` with no couch call. A startup init is awkward here because the audit db name is derived per source db (`auditDbFor(db)`) and the proxy can serve multiple databases, so we'\''d have to enumerate/know them all up front and order it before change processing. The lazy + cached approach handles arbitrary db names with no startup coupling, for negligible cost. Happy to switch to a startup init if we commit to a single, known db — otherwise I'\''d lean on keeping this.

src/audit/audit.service.ts
Abhinegi2: `!existingDoc` is the create case (no prior state) — not deletes. A delete carries its `existingDoc` (the doc being removed), so `needsBaseline` proceeds and, if it'\''s the first audit for that entity, we write a `baseline` (full snapshot from `existingDoc`) *before* the delete record — so restore-from-delete is covered. Entities that already have audit history don'\''t need another baseline since the prior state is replayable. The only no-baseline delete is deleting an entity we never had a stored state for. Adding a unit test pinning the "first audited op is a delete → baseline" case.

src/restricted-endpoints/document/document-write.service.ts
Abhinegi2: actually, it was already in documentController (the PUT field-level permission logic). I just moved the PUT/DELETE write paths into DocumentWriteService, so applyPermission came along.

sleidig: oh, I guess this was partially implemented back then ... :sweat_smile: 
further research seems to confirm this. 

Cool, then let's keep it within the migration :+1:

sleidig (approved):
Nice work! :-)

PR #262, opened 2026-06-12
test: add e2e test suite with mocked CouchDB

sleidig:
## Summary

Adds a real end-to-end test suite (the previous `test/app.e2e-spec.ts` was non-functional NestJS boilerplate): the full `AppModule` is booted against an **in-process mock CouchDB** (`test/utils/mock-couchdb.ts`) that implements the subset of the CouchDB HTTP API used by this backend — `_session`, `_changes` (incl. held longpoll feeds), `_all_docs`, `_bulk_get`, `_bulk_docs`, `_find`, `_local_docs` and document CRUD — and records all upstream requests for assertions.

**40 tests** lock in current behavior of:
- session login (body, basic and JWT-cookie auth), anonymous access via `public` rules
- document CRUD permission enforcement, including condition rules with `${user.name}` variable injection
- `_changes` permission filtering, `lostPermissions`, limit/since handling, tombstones, ignored `_design/` prefixes
- `_bulk_get` / `_all_docs` / `_find` response filtering and `_bulk_docs` request filtering (asserting what is actually forwarded to CouchDB)
- attachment endpoints incl. credential replacement by the proxy
- `/admin/clear_local` checkpoint deletion and **live permission-config updates via the changes feed** (rules swap + automatic clearLocal)

## Motivation

Safety net for the upcoming performance/stability improvements (#261) and response streaming (#109), which will be submitted as separate PRs based on this branch.

## Test plan

- `npm run test:e2e` → 4 suites, 40 tests green
- `npm test` (unit) → unchanged, 174 tests green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end suites covering admin behavior, replication/bulk routes, session + document CRUD with permission filtering, and attachment read/upload/delete flows.
  * Introduced a mock CouchDB server and a shared E2E test bootstrap to seed data and manage app lifecycle.
  * Removed the legacy app E2E test suite.

* **Chores**
  * Increased E2E Jest timeout.
  * Updated CI to run E2E tests as part of the release gate.

Issue #261, opened 2026-06-11
Improve performance, stability and memory use for large databases


sleidig:
## Background

On systems with very large databases the replication backend suffers from high memory use and avoidable load on CouchDB. A code review identified a set of improvements that are independent of response streaming (which is tracked separately in #109).

This issue collects those improvements with a detailed implementation plan.

---

## 1. Internal `_changes` feed permanently fetches full document bodies

`DocumentChangesService` maintains a singleton longpoll `_changes` feed with `include_docs: true` (`src/couchdb/document-changes.service.ts`). Every document write in the whole database flows through the backend with its full body, 24/7. Its only two consumers don't actually need bodies:

- `UserIdentityService.watchUserEntityChanges` only reads `change.id` to invalidate a cache entry (the entity is re-fetched on demand later anyway).
- `RulesService.watchPermissionChanges` only needs the body of the single `Config:Permissions` document.

> **Note:** the client-facing `_changes` endpoint (`ChangesController`) is unaffected — it makes its own direct CouchDB requests with `include_docs: true` and must keep doing so, because permission rules with conditions are evaluated against doc bodies.

**Plan:**
- Remove `include_docs: true` from the internal feed.
- Narrow the emitted type to `{ id, seq, deleted? }` so the compiler enforces that no consumer relies on doc bodies.
- `RulesService` fetches `Config:Permissions` on demand when its ID appears in the feed (bonus: always gets the latest revision instead of a possibly stale intermediate rev from the feed).

**Effect:** converts constant `O(write volume × doc size)` traffic and parsing into near-zero steady-state overhead.

## 2. CASL abilities are rebuilt on every request

`PermissionService.getAbilityFor` runs per HTTP request: `RulesService.injectUserVariablesIntoRules` deep-clones all rules via `JSON.parse(JSON.stringify(...))` with a reviver, then `new Ability(...)` compiles them. The result only changes when the permission config or the user's identity changes — both already have invalidation signals (`UserIdentityService.clearCache()`, the changes feed).

**Plan:**
- Cache `DocumentAbility` per user, keyed by user name + roles, guarded by a permission-config version counter that increments whenever `RulesService` swaps in a new config.
- TTL aligned with the existing `UserIdentityService` cache (5 min).
- Cleared in the same places as the identity cache.

## 3. Basic auth hits CouchDB `_session` on every request

[...]

PR #263, opened 2026-06-12
perf: stop fetching doc bodies on the internal changes feed

sleidig:
## Summary

Part of #261 (item 1). Based on #262 (e2e test suite) — merge that first; GitHub will retarget this PR to master automatically.

The shared longpoll `_changes` feed (`DocumentChangesService`) requested `include_docs=true`, so **every document written to the database was transferred to and parsed by the backend, continuously** — although its only two consumers just compare document IDs:

- `UserIdentityService` only invalidates cache entries by `change.id`
- `RulesService` only needs the body of the single `Config:Permissions` document

> The client-facing `_changes` endpoint (`ChangesController`) is **not** affected: it makes its own direct CouchDB requests with `include_docs=true`, which is required to evaluate permission condition rules against doc bodies.

## Changes

- `DocumentChangesService` no longer requests `include_docs` and emits a narrowed `DocumentChangeEvent { id, seq, deleted? }` — the compiler now enforces that no consumer relies on doc bodies.
- `RulesService` fetches `Config:Permissions` on demand when its ID appears in the feed:
  - serialized via `concatMap` so rapid consecutive config updates cannot apply out of order
  - fetch failures keep the previous in-memory permissions (logged as warning)
  - bonus: the on-demand fetch always returns the latest revision instead of a possibly stale rev from the feed

## Effect

Removes constant `O(write volume × doc size)` transfer/parsing overhead on systems with large, active databases.

## Test plan

- unit tests updated for the new contract (`document-changes.service.spec`, `rules.service.spec`, `user-identity.service.spec`) — 174 green
- e2e suite (incl. live permission-update propagation test, which exercises the new on-demand fetch path end-to-end) — 40 green

## Summary by CodeRabbit

* **New Features**
  * Change-feed notifications are now lightweight, emitting only document ID, sequence, and optional deletion status.
  * Permission rule updates now react to relevant feed events by fetching the latest permission document on demand.
  * Added a public `configVersion` indicator for permission changes.

* **Bug Fixes**
  * Reduced unnecessary document data loading from the change feed.
  * Improved handling for permission deletion, bootstrap/fails-closed behavior, retry/backoff, and update failure recovery.
  * Cache invalidation and local refresh now trigger more reliably when permissions change.


sleidig:
@copilot resolve the merge conflicts in this pull request

PR #267, opened 2026-06-12
perf: efficiency fixes in replication bulk endpoints

sleidig:
## Summary

Part of #261 (items 5 + 6). Based on #262 (e2e suite) — merge that first; GitHub retargets automatically. Independent of the other improvement PRs.

Three small, self-contained fixes in the replication hot path:

1. **`filterBulkDocsRequest` was O(n²)** — `response.rows.find(...)` inside a `.filter()` over the same docs (~10⁶ comparisons for a 1000-doc `_bulk_docs` push). Existing docs are now indexed in a `Map` once.
2. **`_changes` accumulated doc bodies needlessly** — when the client did not request `include_docs`, full doc bodies were still collected across all fetch iterations (potentially tens of MB) and only stripped at the end via a second full mapping pass. Bodies are now dropped per iteration.
3. **`_all_docs` crashed on unknown keys** — CouchDB returns `{key, error: "not_found"}` rows *without an `id`* for missing keys; `isReplicable(row.id)` then threw a `TypeError`, turning the whole request into a 500. Error rows are now passed through (consistent with how `_bulk_get` already handles error entries).

## Test plan

- new unit test: error rows in `filterAllDocsResponse` pass through unchanged
- new e2e test: `POST /:db/_all_docs` with an unknown key returns the `not_found` row instead of a 500
- existing `_bulk_docs` / `_changes` unit + e2e tests cover the refactored paths
- 175 unit / 41 e2e tests green

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved replication responses for unknown document keys by preserving CouchDB error rows instead of filtering them out.
  - Ensured `_all_docs` requests with `include_docs=true` retain expected `not_found` entries for missing documents.
  - Streamlined processing of change results when document bodies are not requested.

- **Tests**
  - Added unit and end-to-end coverage for missing-document error handling.

### DIFF
--- a/src/restricted-endpoints/replication/bulk-document/bulk-document.service.spec.ts
+++ b/src/restricted-endpoints/replication/bulk-document/bulk-document.service.spec.ts
@@ -100,6 +100,18 @@ describe('BulkDocumentService', () => {
     expect(result).toEqual(createAllDocsResponse(childDoc));
   });
 
+  it('should pass through error rows (missing keys) in AllDocs without crashing', () => {
+    const allDocsResponse = createAllDocsResponse(schoolDoc);
+    // CouchDB returns rows without an `id` for unknown keys
+    const errorRow = { key: 'School:missing', error: 'not_found' } as any;
+    allDocsResponse.rows.push(errorRow);
+
+    const result = service.filterAllDocsResponse(allDocsResponse, normalUser);
+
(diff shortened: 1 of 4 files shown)


sleidig:
@coderabbitai review

sleidig:
@coderabbitai review

PR #268, opened 2026-06-12
perf: paginate and throttle clearLocal checkpoint deletion

sleidig:
## Summary

Part of #261 (item 7). Based on #262 (e2e suite) — merge that first; GitHub retargets automatically. Independent of the other improvement PRs.

`clearLocal` (triggered on every permission change and via `/admin/clear_local/:db`) fetched **all** `_local` docs in one buffered response — one replication checkpoint doc exists per client, so large deployments have thousands — and fired **every DELETE in parallel** via `Promise.all`. That momentarily floods CouchDB, and a single failure rejected the whole operation while the remaining deletes kept running detached.

## Changes

- `_local_docs` is fetched in pages of 500; since deleted docs disappear from the index, re-fetching naturally pages through the remainder (an attempted-ids set guards against endless loops on undeletable docs).
- Deletes run with bounded concurrency (10 in flight).
- Failures are logged individually, all docs are still attempted, and a summary error is thrown at the end so the admin endpoint still reports failure.

## Test plan

- new unit tests: internal-doc skipping, bounded concurrency (max in-flight ≤ 10 across 25 docs), failure handling (all attempted, no endless retry, summary error)
- e2e `clear_local` tests unchanged and green
- 177 unit / 40 e2e tests green

### DIFF
--- a/src/admin/admin.service.spec.ts
+++ b/src/admin/admin.service.spec.ts
@@ -1,7 +1,7 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { AdminService } from './admin.service';
 import { CouchdbService } from '../couchdb/couchdb.service';
-import { of } from 'rxjs';
+import { map, of, throwError, timer } from 'rxjs';
 
 describe('AdminService', () => {
   let service: AdminService;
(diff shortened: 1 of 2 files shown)


sleidig:
Maybe not worth the extra complexity at the moment.

PR #269, opened 2026-06-12
perf: compress JSON responses to clients

sleidig:
## Summary

Part of #261 (item 8). Based on #262 (e2e suite) — merge that first; GitHub retargets automatically. Independent of the other improvement PRs.

Responses to clients were never compressed — notably the large JSON payloads of `_bulk_get` / `_all_docs` / `_changes` during initial replication, which matters for clients on slow links.

## Changes

- `compression()` middleware registered in `RestrictedEndpointsModule` (covered by e2e, applies to all routes).
- The middleware's default filter handles the edge cases: responses that already carry a `Content-Encoding` (proxied CouchDB responses), non-compressible content types (attachment binary data), and bodies below the 1 kb threshold are passed through unchanged.
- `compression` streams per chunk (with flush support), so this stays compatible with the planned response streaming (#109).

## Test plan

- new e2e spec: large JSON gzipped when accepted, identity respected, small responses untouched, replication endpoint responses compressed
- 174 unit / 44 e2e tests green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added automatic gzip compression for large JSON responses when supported by the client.
  * Responses that are small, already encoded, or explicitly request identity encoding remain uncompressed.
  * Replication endpoint responses now support compression.

PR #270, opened 2026-06-12
feat: stream replication responses instead of buffering them (#109)

### LINKED ISSUE #109: Stream instead of buffer results
Normal `async`/`await` NodeJS fetch requires to load all data of a request and buffer it.
Using NodeJS streams its possible to read the request line by line and directly forward the results. 
The reduces the amount of data that needs to be buffered.
The difficulty is that the results cannot be parsed as plain JSON but need to be adjusted.

sleidig:
## Summary

Implements #109 (plan: see [this comment](https://github.com/Aam-Digital/replication-backend/issues/109#issuecomment-4678485577)) for the read-heavy replication endpoints: `_all_docs` (GET/POST), `_bulk_get`, `_find` and `_changes`.

Based on #262 (e2e suite) — merge that first; GitHub retargets automatically.

## Problem

Every proxied response was fully buffered by axios, `JSON.parse`d into an object graph, filtered into a copy, and re-serialized as one giant string — a peak memory cost of roughly **4–5× the raw payload per request**, multiplied by concurrent syncing clients. An unbounded `GET /db/_all_docs?include_docs=true` could pull the entire database into heap.

## Changes

- **`CouchdbService.getStream`/`postStream`** — axios `responseType: 'stream'`; `Accept-Encoding: identity` on the internal hop (axios doesn't decompress streams; client-facing gzip comes from the compression middleware, #269); upstream error responses are read and mapped to the same `HttpException`s as the buffered methods.
- **`JsonArrayFilterTransform`** (`src/common/json-array-filter.ts`) — incrementally re-serializes a JSON object from `stream-json` tokens, filtering/transforming the items of one top-level array while passing envelope fields through (before *and* after the array, e.g. `total_rows` / `bookmark`) — without ever holding the document or the array in memory. 12 dedicated unit tests (filtering, transformation, escaping, chunked input, truncated/malformed input, non-object roots).
- **`_all_docs` / `_bulk_get` / `_find`** pipe the CouchDB response through this filter directly into the client response via `stream.pipeline` (full backpressure handling). Per-item permission predicates are shared between the streaming endpoints and the existing buffered service methods.
- **`_changes`** keeps its iterative fetch loop and 8 s time budget, but writes each batch's permitted results to the client immediately instead of accumulating everything; the envelope (`last_seq`, `pending`, `lostPermissions`) is appended at the end. First bytes flow early, which also reduces browser idle-timeout pressure.

## Error semantics

- Upstream failure **before the first byte** → regular error response (status preserved), e.g. unknown db.
- Failure **mid-stream** → the connection is aborted (`res.destroy()`) so clients see a truncated response and retry, instead of mistaking partial JSON for a complete result. Truncated *upstream* bodies are detected by the transform (stream errors on unexpected end) and can never produce complete-looking output. Covered by a dedicated e2e test that cuts the mock-CouchDB connection mid-body.

## Behavior changes

- Streamed POST endpoints (`_bulk_get`, `_all_docs`, `_find`) now return **200** (matching real CouchDB) instead of NestJS's default 201. PouchDB accepts both.
- `lostPermissions` moves to the end of the `_changes` JSON object (key order is irrelevant to JSON parsers).
- `stream-json@1.x` added as dependency (CJS; v3 is ESM-only).

## Note on sibling PRs

`bulk-document.service.ts` is also touched by #267 — the per-row filter here already includes #267's `not_found`-row guard, so whichever merges second needs a trivial rebase.

Possible follow-up (out of scope): the design-document view query endpoint also buffers potentially large responses and could reuse the same utility.

## Test plan

- 189 unit / 41 e2e tests green (the whole e2e suite from #262 now runs against the streamed implementations — output equivalence is verified end-to-end)
- new unit tests: transform (12), streamed controller endpoints incl. abort/rethrow paths (8), changes controller adapted to streamed responses (22)
- new e2e test: mid-stream upstream failure → aborted connection

## Summary by CodeRabbit

* **New Features**
  * Replication endpoints now stream large CouchDB responses directly to clients, improving responsiveness and reducing memory usage.
  * Added streamed JSON filtering to transform or omit permitted items as data is processed.
* **Bug Fixes**
  * Truncated, malformed, or interrupted CouchDB responses now fail cleanly instead of returning partial data.
  * Improved handling of upstream errors and standardized success status codes.
* **Tests**
  * Expanded coverage for streaming, permission filtering, backpressure, disconnects, and transfer failures.


sleidig:
@copilot resolve the merge conflicts in this pull request and run lint formatting

Abhinegi2 (approved):

PR #278, opened 2026-06-23
feat(audit): expose audit-enabled status at GET /_features/audit

Abhinegi2:
## What
Adds a small public endpoint **`GET /_features/audit`** → `{ "enabled": <AUDIT_ENABLED> }` so the frontend (ndb-core change-history UI) can show/hide the feature based on whether change logging is on.

## Why (replaces aam-services#142)
We first added the flag to aam-backend-service's `/actuator/features`, but @sleidig rightly flagged the trade-off: that service would advertise a feature it doesn't implement, and the flag had to be **kept in sync by hand** with this repo's `AUDIT_ENABLED` (which already bit us on dev — recording was on but the UI stayed disabled until a separate, untracked env was set). So aam-services#142 was closed and the status is reported **here**, where the flag lives — single source of truth, no sync.

## How
- Factor `isAuditEnabled(config)` and reuse it for **both** the `AuditModule` service-selection factory and the new endpoint, so the reported state can't drift from the actual recording behavior. (It also fixes a latent bug: `config.get<boolean>` returned the raw string, so `AUDIT_ENABLED=false` previously read as truthy.)
- `AuditFeatureController` is registered in `AuditModule`, which is imported **before** the CouchDB-proxy `:db` routes in `AppModule`, so the literal path wins route matching (isn't captured by `/:db/:docId`). Public — it returns only a boolean feature flag.

## Tests
- Unit: controller returns enabled true/false for `true` / `'true'` / `'false'` / unset.
- e2e: unauthenticated `GET /_features/audit` returns the flag through the full app (pins route precedence + public access).

## Follow-up (frontend)
ndb-core change-history check will point at `DB_PROXY_PREFIX + '/_features/audit'` instead of `/actuator/features`; dev's manual `FEATURES_AUDIT_ENABLED` env then gets removed.


src/audit/audit-feature.controller.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  return new AuditFeatureController(config);
+}
+
+it('reports enabled=true when AUDIT_ENABLED is true (boolean)', () => {
+  expect(controllerWith(true).getAuditFeature()).toEqual({ enabled: true });
+});
+
+it("reports enabled=true when AUDIT_ENABLED is the string 'true'", () => {
+  expect(controllerWith('true').getAuditFeature()).toEqual({ enabled: true });
+});
```
sleidig: I think the distinction between string/boolean should not be in a controller but further down, in the config service (which it already is - let's only test it at that level then)

src/audit/audit-feature.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+ * `/:db` routes in {@link AppModule}, so this literal path wins route matching
+ * and is not captured by `/:db/:docId`. No authentication is required: it
+ * returns only a non-sensitive boolean feature flag (the auth middleware passes
+ * unauthenticated requests through for routes without `@OnlyAuthenticated()`).
+ */
+@Controller()
+export class AuditFeatureController {
+  constructor(private readonly config: ConfigService) {}
+
+  @Get('_features/audit')
```
sleidig: I am thinking about the best, most extendable pattern for the API here in general:

Maybe `/<module>/_status` (--> `/audit/_status`) could be a good pattern? (or is there a standard route for health / status check endpoints commonly used? Let's briefly research)

aam-services has a single `/actuator/features` endpoint across all modules. That could be an alternative (I would just use `/features` then here). Then we would need a separate, central features controller and a service that allows modules to register themselves (or maybe a multi provider through DI).

I am not sure what is more elegant. Do you have thoughts, @Abhinegi2 ?

Abhinegi2: @sleidig  first, I was thinking of reusing /features, but the CouchDB proxy  `/:db `matches any first segment, so /features would grab the db name features. Right now, no such DB exists, but if one ever does, it'd collide.

CouchDB forbids db names starting with `_` and already uses that prefix for its own meta routes (_session, _changes, _all_docs…), so `_features` can never collide(that's my thought)

So my approach is to go central like actuator: one `GET /_features` → `{ "audit": { "enabled": true } }`. New flags become new keys, no new routes.

src/features/features.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+ * without `@OnlyAuthenticated()`).
+ */
+@Controller()
+export class FeaturesController {
+  constructor(private readonly config: ConfigService) {}
+
+  @Get('_features')
+  getFeatures(): { audit: { enabled: boolean } } {
+    return {
+      audit: { enabled: isAuditEnabled(this.config) },
```
sleidig, line 33: we could decouple this completely, creating a FeatureService with a `registerFeature` method that each module calls in their setup. But probably over-engineered for now. Let's keep this as is.

sleidig (approved):
Code looks good 👍

PR #279, opened 2026-06-27
feat: allow creation of indices for authenticated users

TheSlimvReal:
Allows authenticated users to create indices for find queries (see [docs](https://docs.couchdb.org/en/stable/api/database/find.html#post--db-_index)).
This is needed for the [database pagination feature](https://github.com/Aam-Digital/ndb-core/pull/4046), especially for using `sort` inside a mango query an index needs to be available.

Users are only enforced to be logged in, no further permission checks are required.
The idea is that the indices are lazy-generated, meaning only when they are first needed.

Live on [https://load-test.aam-digital.net/](https://load-test.aam-digital.net/c/child)

## Summary by CodeRabbit

* **New Features**
  * Added a new authenticated endpoint to create a database index for bulk document operations.
  * Requests to the new endpoint now forward the submitted payload directly to the backend service.

* **Security**
  * Restricted the new index-creation action so it is available only to signed-in users.

### DIFF
--- a/src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
+++ b/src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
@@ -21,6 +21,7 @@ import {
   FindResponse,
 } from './couchdb-dtos/bulk-docs.dto';
 import { BulkGetRequest, BulkGetResponse } from './couchdb-dtos/bulk-get.dto';
+import { OnlyAuthenticated } from '../../../auth/only-authenticated.decorator';
 
 /**
  * Handle endpoints for the CouchDB replication process and bulk actions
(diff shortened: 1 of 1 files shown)


sleidig:
## How does this relate to the existing `_design` document controller?

Both endpoints are needed — they serve **two different CouchDB query systems**:

| | `PUT /:db/_design/:designName` (existing) | `POST /:db/_index` (this PR) |
|---|---|---|
| Query system | **MapReduce views** | **Mango / `_find`** |
| You write | JS `map`/`reduce` functions | A declarative JSON index definition (`{index: {fields: [...]}}`) |
| Queried by | `GET .../_view/:viewName` | `POST /:db/_find` with a selector + `sort` |
| Purpose | Pre-computed, permission-filtered views | Speeds up `_find` queries; **required for `sort`** |

CouchDB has two independent ways to query data:

1. **Views** (`_design` docs) — the classic approach with map/reduce functions. The existing `DesignDocumentController` creates these and queries them with read-permission filtering.
2. **Mango queries** (`_find`) — a declarative, MongoDB-style JSON query language. To make a `_find` query efficient (and *mandatory* for sorting), CouchDB needs a Mango **index**, created via `POST /:db/_index` — which is what this PR adds for the [database pagination feature (#4046)](https://github.com/Aam-Digital/ndb-core/pull/4046).

So they're complementary, not redundant.

### One thing worth flagging

Under the hood, a Mango `_index` is stored as a `_design/...` document (with `"language": "query"`). So there's now partial overlap in responsibility, with **different authorization** on the two paths:

- `_index` (this PR): only requires authentication, no permission check.
- `putDesignDoc`: requires the `manage` / `_design` permission.

That's a defensible, intentional split (lazy-generated, logged-in-only), but it does mean design documents can be created through two paths with different authz. Worth a conscious decision rather than leaving it implicit.

sleidig (approved):
Looks good to me 👍

PR #266, opened 2026-06-12
fix: add request timeout and keep-alive socket cap for CouchDB requests

sleidig:
## Summary

Part of #261 (item 4). Based on #262 (e2e suite) — merge that first; GitHub retargets automatically. Independent of the other improvement PRs.

Two stability gaps in the CouchDB HTTP client:

- **No request timeout anywhere** — a hung CouchDB connection (half-open socket, stalled network) hangs the client request forever. The internal longpoll changes feed could stall silently too: its `timeout=50000` parameter only controls the *server-side* longpoll window, not the client socket.
- **No explicit keep-alive agent with a socket cap** — request bursts (many syncing clients) could open an unbounded number of connections to CouchDB.

## Changes

- `CouchdbModule` now registers `HttpModule` with an exported, unit-tested factory (`couchdbHttpOptions`):
  - `timeout`: default **60 s** (`DATABASE_TIMEOUT_MS`) — safely above the 50 s longpoll window, so feed requests are never aborted client-side while idle
  - `httpAgent`/`httpsAgent`: keep-alive, default **50 sockets max** (`DATABASE_MAX_SOCKETS`)
- New env vars documented in README and `.env.template`; invalid values fall back to defaults.

When a timeout strikes the changes feed, the existing retry-with-backoff logic in `DocumentChangesService` takes over — previously the feed would just hang with no recovery.

## Test plan

- new unit tests for the options factory (defaults, env overrides, invalid value fallback)
- full suite: 177 unit / 40 e2e green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added optional CouchDB connection settings for request timeouts and maximum keep-alive sockets.
  * CouchDB connections now reuse sockets and apply configurable connection limits.

* **Bug Fixes**
  * Prevented excessively short timeouts from interrupting long-running feed requests.
  * Invalid connection settings now safely fall back to defaults.

* **Documentation**
  * Updated setup documentation with the new optional environment variables.

* **Tests**
  * Added coverage for defaults, overrides, invalid values, socket limits, and timeout safeguards.

PR #280, opened 2026-06-27
combined branch for Improve performance, stability and memory use for large databases

### LINKED ISSUE #261: Improve performance, stability and memory use for large databases

Abhinegi2:
Combines all PRs from the #261 perf stack onto one branch so the integrated behavior can be tested in CI and on staging (Tier-3 real-replication / memory test). The actual changes merge via the individual PRs #263–#270.

- Unit 231 passed, e2e 50 passed, lint clean.
- Manual Tier-2 vs real CouchDB 3.5.2: permission filtering, _changes doc-strip + lostPermissions, not_found passthrough, streamed 200 reads, gzip compression, clearLocal, timeout fail-fast.
- Observed: brief fail-closed window if a request hits before `Config:Permissions` finishes loading at startup (self-corrects); likely pre-existing.

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable CouchDB timeout and connection limits.
  * Improved performance with cached login and permission checks.
  * Large CouchDB responses can now be streamed and compressed more efficiently.

* **Bug Fixes**
  * Reduced failures during bulk document and changes requests by handling large responses incrementally.
  * Improved stability when deleting local documents and processing partial failures.
  * Change notifications now use a lighter payload and better tolerate retries and missing data.


Abhinegi2:
## Manual UI testing steps ([dev site](https://dev.aam-digital.net/))

### A. Initial sync / load — streaming (#270, #263, #267)
- [x] Fresh login (incognito / cleared site data) → full initial replication: all lists populate, **no console/network errors**, completes in reasonable time
- [x] Reload app → fast incremental sync (`_changes` since last_seq), data intact

### B. Permission filtering — must not regress
- [x] Limited-role user sees only permitted entity types/records
- [x] Admin sees everything
- [x] Revoke a permission / change a role → next sync purges now-forbidden records locally (lostPermissions)

### C. Write path — `_bulk_docs` + audit (#267)
- [x] Create a record → saves, appears, persists after reload (synced to server)
- [x] Edit a record → change persists + syncs
- [x] Delete a record → removed + syncs (tombstone)
- [x] Limited user edits a type they cannot write → denied gracefully (no crash)

### D. Bulk / large data (item 5, streaming)
- [x] CSV / bulk import of many records → all imported, no timeout/hang

### E. Resilience — auth cache (#265), timeouts (#266)
- [x] Log out / log in repeatedly → login always works (60s auth cache)
- [ ] App stays responsive, no infinite spinners

george-neha:
1. Under Recurring Activities- the participant list is blank for some records  check for record Class 6B
Screenshot 2026-06-29 at 11 23 22 AM
2. There is mismatch in the participant list shown under recurring Attendance and Record Attendance- check for driving class. under record attendance, attendance is being captured for 36 participants
image
3. In import, in step 4- review and edit data, the data takes some time to load
image

Abhinegi2:
@george-neha 
1. I checked this issue, and it seems to be caused by a corrupted database record, which is why the participant list is broken for that recurring activity. I also created a new recurring activity and couldn't reproduce the issue there, so it doesn't appear to affect newly created recurring activities.

2. I checked this as well. It looks like there are two different records named "Driving Class"
one belongs to the Recurring Activity entity and the other to the Event entity. To verify this, I renamed the Event record from "Driving Class" to "Driving Class 1", and the difference became visible. Since this is the dev site, it appears to be a data inconsistency rather than an application issue.

Abhinegi2:
also investigated the import loading issue and found that it wasn't actually a bug. the csv I was importing contained the Address field, and while mapping the columns, I didn't check the `Skip Address Lookup` checkbox. Because of that, the import was taking 2–3 minutes, depending on the number of rows. Once I enabled that option, the import completed much faster. 
 I think this wasn't an actual issue after all. @george-neha, could you confirm by checking with your sheet as well?

sleidig:
@Abhinegi2 , sorry for being picky :blush:   But could you apply the cleanups back to the individual branches, please? Replication backend is mission-critical and clear, isolated git commits are super helpful in case we need to troubleshoot regressions or undo any changes later.

PR #289, opened 2026-07-24
fix: reserved permission keys can no longer collide with Keycloak role names

### LINKED ISSUE #236: Reserved role names (`default`, `public`) collide with Keycloak role names in getRulesForUser
## Problem

In `RulesService.getRulesForUser` (`src/permissions/rules/rules.service.ts`) the lookup of rules for a user is:

```ts
const userRules = user.roles
  .filter((role) => this.permission.hasOwnProperty(role))
  .map((role) => this.permission[role])
  .flat();
if (this.permission.default) {
  userRules.unshift(...this.permission.default);
}
```

The `RulesConfig` document uses two reserved keys with special semantics:

- `default` — rules prepended to **every** authenticated user
- `public` — rules used when the user is **not** logged in

Because the role lookup is `this.permission.hasOwnProperty(role)` against the same object that holds those reserved keys, a Keycloak role literally named `default` or `public` will be picked up as if it were a normal role mapping. In practice that means:

- A user with a Keycloak role `default` would unintentionally re-apply (or duplicate) the `default` rules.
- A user with a Keycloak role `public` would receive the unauthenticated `public` rules in addition to their normal rules — which can be broader than intended (the `public` block is typically used to allow the login page to read minimal config without authentication).

This is a misconfiguration foot-gun, not an exploitable bypass on its own, but it can lead to access being granted that an admin did not intend.

## Suggested fix

- Explicitly skip the reserved keys in the role lookup, e.g. `user.roles.filter(role => role !== 'default' && role !== 'public' && this.permission.hasOwnProperty(role))`.
- Optionally validate the permission document on load and warn if any non-reserved key collides with a known reserved name in the future.
- Add unit tests for both reserved names.

## Related

Came up while reviewing permission-evaluation paths around the v1.6.0–v1.6.5 releases (see also #235 / fail-closed startup hardening).

Abhinegi2:
closes #236

## What this does

Reserved permission-config sections could collide with Keycloak role names. `Config:Permissions` uses two keys with special meaning, `default` (prepended to every authenticated user) and `public` (used for anonymous visitors). The role lookup resolved a user's roles as keys of the same object, so a Keycloak role literally named `default` or `public` was picked up as a role, duplicating the default rules or granting the anonymous rules to a logged-in user.

Changes:
- Rename the reserved sections to `_default` / `_public`. The underscore marks them internal so they can never collide with a realm role name.
- The rules lookup now skips any role starting with `_` and any reserved section name.
- Reads accept both the new and the legacy spellings (prefers new), so this can deploy before any document is migrated.
- The managed-defaults write-back maintains `_default` and migrates a legacy `default` section across on its next write.

Backward compatible: existing documents keep working via the legacy-key fallback. No Keycloak changes (the reserved names are config keys, not realm roles).

## Manual Testing Steps

Verified on a local dev stack (CouchDB + Keycloak + this backend built from the branch):

1. Create a Keycloak realm role literally named `public` and one named `_something`, assign both to a normal test user.
2. Sign in as that user: neither injects the reserved `public`/`_` section into the user's rules (the user only gets their real role's rules plus the `_default` rules).
3. On an instance whose `Config:Permissions` still uses the legacy `default`/`public` keys, confirm the app and backend still resolve permissions correctly (legacy fallback), then confirm the managed-defaults write-back migrates the `default` section to `_default`.

## Summary by CodeRabbit

* **New Features**
  * Added support for renamed permission sections (`_public`/`_default`) with automatic compatibility for legacy `public`/`default` keys.
  * Permission changes now expose an updated change-notification stream and use more robust, limited on-demand refresh behavior.
* **Bug Fixes**
  * Public and default access rules are now consistently resolved, including correct precedence when both legacy and updated sections exist.
  * Reserved configuration keys and underscore-prefixed role names are excluded from role resolution to prevent unintended access.


Abhinegi2:
@coderabbitai review

src/permissions/rules/permission.ts
```diff
@@ -1,13 +1,43 @@
...
- * The rules defined in 'default' will be prepended to any other rules defined for a user
- * The rules defined in 'public' will be used if a user is not logged in
+ * The rules defined in '_default' (legacy 'default') are prepended to any other rules defined for a user.
+ * The rules defined in '_public' (legacy 'public') are used if a user is not logged in.
  */
 export type RulesConfig = {
+  _public?: DocumentRule[];
+  _default?: DocumentRule[];
   public?: DocumentRule[];
   default?: DocumentRule[];
```
sleidig, line 57: mark `// @deprecated`?

src/permissions/rules/rules.service.ts
```diff
@@ -263,16 +273,17 @@ export class RulesService implements OnModuleInit {
...
       return 'done';
     }
-    const { merged, changed, dropped } = mergeManagedDefaults(
-      doc.data.default,
-    );
-    if (!changed) {
+    const currentDefault =
+      doc.data[DEFAULT_SECTION_KEY] ?? doc.data[LEGACY_DEFAULT_KEY];
+    const hasLegacyDefault = doc.data[LEGACY_DEFAULT_KEY] !== undefined;
+    const { merged, changed, dropped } = mergeManagedDefaults(currentDefault);
```
sleidig, line 407: let's create a issue for removing backwards compat / breaking changes of the next major version, where we will remove this again to clean up the code base.

src/permissions/rules/rules.service.ts
```diff
@@ -318,18 +329,30 @@ export class RulesService implements OnModuleInit {
    */
   getRulesForUser(user: UserInfo): DocumentRule[] {
     if (!user) {
-      return this.permission?.public ?? [];
+      return (
+        this.permission?.[PUBLIC_SECTION_KEY] ??
+        this.permission?.[LEGACY_PUBLIC_KEY] ??
```
sleidig: could we merge this before assigning it to this.permissions and then not need the fallbacks across the class here?

sleidig (approved):
Looks good overall

src/permissions/rules/rules.service.ts
Abhinegi2: done -https://github.com/Aam-Digital/replication-backend/issues/295

sleidig (approved):

sleidig:
@Abhinegi2 , I have not tested this functionally again. But if you have, this should be good to go :-)

PR #294, opened 2026-07-29
fix(sentry): fingerprint HttpExceptions so CouchDB faults stop merging into one issue

### LINKED ISSUE #293: REPLICATION-BACKEND-A7: all CouchDB 5xx errors collapse into one opaque "HttpException: Http Exception" issue
Filed by the automated monitoring routine. Window: **2026-07-28 02:16 CEST → 2026-07-29 20:30 CEST**.

## Summary

[REPLICATION-BACKEND-A7](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-A7) — `HttpException: Http Exception` — is **escalating**, with 51 events in this window and **828 occurrences / 18 users** lifetime (first seen 2026-06-11).

The issue is not actionable as it stands: its title carries **neither the HTTP status nor the failing route**, so every distinct CouchDB failure across every endpoint is grouped into a single issue. This is the mirror image of the usual fingerprinting problem — instead of one fault splitting across many issue IDs, many different faults are merging into one.

## Root cause of the grouping

`src/couchdb/couchdb.service.ts:55`:

```ts
private initMapAxiosErrorsToNestjsExceptions() {
  this.httpService.axiosRef.interceptors.response.use(undefined, (err) => {
    const resultErr = err.response
      ? new HttpException(err.response.data, err.response.status)
      : err;
    return Promise.reject(resultErr);
  });
}
```

Every axios error becomes a bare `HttpException`, whose message renders as the constant string `"Http Exception"`. Sentry groups on that, so status and route are erased.

Note that `beforeSend` in `src/sentry.configuration.ts` already **drops 4xx** `HttpException`s:

```ts
if (error instanceof HttpException && error.getStatus() >= 400 && error.getStatus() < 500) {
  return null;
}
```

so the events that survive into A7 are CouchDB **5xx** responses (inferred from this filter, not from the event payloads — the status is not recoverable from the grouped events, which is exactly the problem).

The same `beforeSend` already fingerprints **log messages** by their static prefix, but does nothing equivalent for exceptions.

## Observed burst (this window)

All 51 events are on **`dev.aam-digital.net`**, `environment: development`, user `User:demo-admin`, release `1.10.2`. Not production, not user-facing.

Timeline: a continuous ~38-minute burst, **2026-07-29T09:37:16Z → 10:15:28Z**, spanning a full replication cycle:

- `POST /app/_bulk_docs`
- `GET /app/_local/6bP4PU_O06.tuLk.V6jAeA==`, `GET /app/_local/nEFc8Uwkl2o_ITqxaq4jOA==`, `GET /app/_local/I0eUfZ0LCznRa1vteCSrQw==` (replication checkpoints)
- `GET /app/_all_docs`
- `GET /app/_design/search_index`

That every stage of one replication fails for one dev user, at a steady ~30s cadence, suggests a single sustained backend condition rather than many independent faults — but the grouping makes that impossible to confirm from Sentry alone.

**Not related to v1.10.9.** The burst predates the v1.10.9 release (2026-07-29 20:05 CEST, #291 "log bulk-docs pushes silently dropped for missing write permission") by ~10 hours, and the events report release `1.10.2`. Flagging this because the titles look superficially related.

## Proposed fix

Extend the existing `beforeSend` to fingerprint `HttpException`s by status and route, so distinct faults separate into distinct issues. This is a Sentry-grouping-only change: the `HttpException` returned to clients is untouched, so no API behaviour changes.

A PR implementing this is linked below.

Once grouping is fixed, the dev burst above should be re-examined — with a status code and route attached it should be immediately clear whether it was one CouchDB fault or several.

## Related

- #286 already tracks the separate `REPLICATION-BACKEND-A` DEP0060 console noise (last run recommended a capture filter for it; that recommendation is already covered there, so it is not duplicated here).

sleidig:
Closes #293

## Problem

`CouchdbService.initMapAxiosErrorsToNestjsExceptions` (`src/couchdb/couchdb.service.ts:55`) maps every failed axios response to a bare `HttpException`:

```ts
const resultErr = err.response
  ? new HttpException(err.response.data, err.response.status)
  : err;
```

That exception's message renders as the constant string `"Http Exception"`. Sentry groups on it, so **every CouchDB fault — any status, any route — merges into a single issue whose title contains neither a status code nor a route.**

That issue is [REPLICATION-BACKEND-A7](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-A7): **828 occurrences / 18 users** lifetime, currently `escalating`. In the window this was found (2026-07-28 → 2026-07-29) it absorbed 51 events spanning `POST /app/_bulk_docs`, `GET /app/_all_docs`, `GET /app/_design/search_index` and three different `_local/*` replication checkpoints — with no way to tell from Sentry which fault was which, or even what status CouchDB returned.

Note the existing `beforeSend` already drops 4xx `HttpException`s, so what lands in A7 is CouchDB **5xx** responses — the ones most worth being able to distinguish.

## Change

Fingerprint `HttpException` events, splitting on whatever actually distinguishes one fault from another:

```ts
const status = error.getStatus();
event.fingerprint = UPSTREAM_AVAILABILITY_STATUSES.has(status)
  ? ['HttpException', String(status)]
  : ['HttpException', String(status), event.transaction ?? UNKNOWN_ROUTE];
```

- **Operation-specific failures** (500, and anything not listed below) group by **status + route**. These genuinely differ per endpoint.
- **Upstream-availability failures** — `502 Bad Gateway`, `503 Service Unavailable`, `504 Gateway Timeout` — group by **status alone**. These describe CouchDB (or a gateway in front of it) being unreachable, which is one condition regardless of which request observed it. Including the route would split a single outage into as many issues as there are endpoints in flight — the same fragmentation this PR exists to prevent, just on a different axis. Thanks to @sleidig for catching this in review.

`event.transaction` is the parameterised Nest route (e.g. `POST /:db/_bulk_docs`), not the raw URL, so document IDs don't fragment groups the other way.

This affects **Sentry grouping only**. The `HttpException` propagated to the client is untouched, so response bodies and status codes are unchanged — there is no API behaviour change.

`beforeSend` is extracted from the inline `Sentry.init` options and exported, because it had no test coverage and its rules decide whether an issue is actionable at all. The existing behaviours (dropping 4xx, grouping log messages via `normalizeLogMessage`) are preserved unchanged and now covered by tests.

## Verification

- Grouping tests cover both directions — under-splitting *and* over-splitting:
  - 4xx dropped; 5xx reported
  - 500 fingerprints by status + route; the same 500 on two different routes stays distinct
  - 502/503/504 on `POST /:db/_bulk_docs` and `GET /:db/_all_docs` produce an **identical** fingerprint
  - different upstream statuses still separate from each other (502 ≠ 503)
  - an operation-specific failure never merges with an upstream outage on the same route (500 ≠ 502)
  - unknown-route fallback; existing message grouping preserved; `HttpException` fingerprint takes precedence over the message; non-`HttpException` errors passed through untouched
- `npx jest` — full suite green: **275 tests / 32 suites**
- `npx nest build` — clean
- `npx prettier --check` — clean

## Open question

I drew the upstream-availability line at 502/503/504 and deliberately left **507 Insufficient Storage** out. It is arguably also route-independent — a full disk is a server-state condition, not a property of the request — but I have not observed CouchDB returning it here, so I would rather not encode a guess. Easy to add if reviewers think it belongs.

## Follow-up

Once this is deployed, the dev burst described in #293 should be re-examined: with a status and route attached it should be immediately clear whether it was one sustained CouchDB fault or several independent ones. Existing A7 events will stay in the old group; new events will split.

### DIFF
--- a/src/sentry.configuration.spec.ts
+++ b/src/sentry.configuration.spec.ts
@@ -1,4 +1,6 @@
-import { normalizeLogMessage } from './sentry.configuration';
+import { HttpException } from '@nestjs/common/exceptions/http.exception';
+import * as Sentry from '@sentry/node';
+import { beforeSend, normalizeLogMessage } from './sentry.configuration';
 
 describe('normalizeLogMessage', () => {
   it('keeps a constant message unchanged', () => {
@@ -42,5 +44,111 @@ describe('normalizeLogMessage', () => {
     const b = normalizeLogMessage('Failed to obtain Keycloak access token');
 
     expect(a).not.toBe(b);
+  });
+});
+
+describe('beforeSend', () => {
+  function event(
+    overrides: Partial<Sentry.ErrorEvent> = {},
+  ): Sentry.ErrorEvent {
+    return { type: undefined, ...overrides } as Sentry.ErrorEvent;
+  }
+
+  function hint(originalException?: unknown): Sentry.EventHint {
+    return { originalException } as Sentry.EventHint;
+  }
+
+  it('drops 4xx HttpExceptions, which are client errors rather than faults', () => {
+    const result = beforeSend(
+      event(),
+      hint(new HttpException('not found', 404)),
+    );
+
+    expect(result).toBeNull();
+  });
+
+  it('reports 5xx HttpExceptions', () => {
+    const result = beforeSend(event(), hint(new HttpException('boom', 500)));
+
+    expect(result).not.toBeNull();
+  });
+
+  it('fingerprints HttpExceptions by status and route so unrelated CouchDB faults do not merge', () => {
+    const result = beforeSend(
+      event({ transaction: 'POST /:db/_bulk_docs' }),
+      hint(new HttpException('boom', 502)),
+    );
+
+    expect(result?.fingerprint).toEqual([
+      'HttpException',
+      '502',
+      'POST /:db/_bulk_docs',
+    ]);
+  });
+
+  it('separates the same status on different routes', () => {
+    const bulkDocs = beforeSend(
+      event({ transaction: 'POST /:db/_bulk_docs' }),
+      hint(new HttpException('boom', 500)),
+    );
+    const allDocs = beforeSend(
+      event({ transaction: 'GET /:db/_all_docs' }),
+      hint(new HttpException('boom', 500)),
+    );
+
+    expect(bulkDocs?.fingerprint).not.toEqual(allDocs?.fingerprint);
+  });
+
+  it('separates different statuses on the same route', () => {
+    const serverError = beforeSend(
+      event({ transaction: 'POST /:db/_bulk_docs' }),
+      hint(new HttpException('boom', 500)),
+    );
+    const badGateway = beforeSend(
+      event({ transaction: 'POST /:db/_bulk_docs' }),
+      hint(new HttpException('boom', 502)),
+    );
+
+    expect(serverError?.fingerprint).not.toEqual(badGateway?.fingerprint);
+  });
+
+  it('falls back to an explicit placeholder when no route is known', () => {
+    const result = beforeSend(event(), hint(new HttpException('boom', 500)));
+
+    expect(result?.fingerprint).toEqual([
+      'HttpException',
+      '500',
+      '<unknown route>',
+    ]);
+  });
+
+  it('fingerprints plain log messages using normalizeLogMessage', () => {
+    const message = 'Changes feed error (failure #1): timeout';
+    const result = beforeSend(event({ message }), hint());
+
+    expect(result?.fingerprint).toEqual([normalizeLogMessage(message)]);
+  });
+
+  it('prefers the status/route fingerprint over the normalized message for HttpExceptions', () => {
+    const result = beforeSend(
+      event({
+        message: 'Request failed: something dynamic',
+        transaction: 'GET /:db/_changes',
+      }),
+      hint(new HttpException('boom', 503)),
+    );
+
+    expect(result?.fingerprint).toEqual([
+      'HttpException',
+      '503',
+      'GET /:db/_changes',
+    ]);
+  });
+
+  it('passes through non-HttpException errors without a message unchanged', () => {
+    const result = beforeSend(event(), hint(new Error('kaboom')));
+
+    expect(result).not.toBeNull();
+    expect(result?.fingerprint).toBeUndefined();
   });
 });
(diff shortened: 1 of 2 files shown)


src/sentry.configuration.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      expect(result).toBeNull();
+    });
+
+    it('reports 5xx HttpExceptions', () => {
+      const result = beforeSend(event(), hint(new HttpException('boom', 500)));
+
+      expect(result).not.toBeNull();
+    });
+
+    it('fingerprints HttpExceptions by status and route so unrelated CouchDB faults do not merge', () => {
```
sleidig: A generic "Bad Gateway" should not be logged differently across different requests. But other errors possibly should be distinct.

sleidig: Good catch — you're right, and the way I had it would have actively made an outage harder to read. A 502 means CouchDB is unreachable; that's one condition, so keying it on route would split a single outage into as many issues as there are endpoints in flight. Same fragmentation this PR exists to prevent, just on a different axis.

Fixed in 0816a95. Statuses that describe *upstream availability* rather than the request now group by status alone:

```ts
const UPSTREAM_AVAILABILITY_STATUSES: ReadonlySet<number> = new Set([
  502, // Bad Gateway
  503, // Service Unavailable
  504, // Gateway Timeout
]);
```

```ts
event.fingerprint = UPSTREAM_AVAILABILITY_STATUSES.has(status)
  ? ['HttpException', String(status)]
  : ['HttpException', String(status), event.transaction ?? UNKNOWN_ROUTE];
```

Everything else — 500 in particular — is operation-specific and keeps its route, which is your "other errors possibly should be distinct."

Tests pin both halves: 502/503/504 on `POST /:db/_bulk_docs` and `GET /:db/_all_docs` produce an identical fingerprint, while 500 on those same two routes stays distinct. Two guards against over-merging in the other direction: different upstream statuses still separate from each other (502 ≠ 503), and an operation-specific failure never merges with an upstream outage on the same route (500 ≠ 502).

**One judgement call worth a second opinion:** I drew the line at 502/503/504 and deliberately left **507 Insufficient Storage** out. It's arguably also route-independent — a full disk is a server-state condition, not a property of the request — but I've not seen CouchDB return it here, so I'd rather not encode a guess. Happy to add it if you think it belongs.

Full suite green after the change: 275 tests / 32 suites, build and prettier clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #310, opened 2026-08-05
fix(sentry): stop reporting client disconnects and DNS blips as faults

sleidig:
Found by the monitoring routine on 2026-08-05. Two sources of Sentry noise that both appeared the moment `1.11.4` reached production, neither of which is a fault of this service.

## 1. Client disconnects reported as an escalating Sentry issue

[REPLICATION-BACKEND-BF](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-BF) — *"aborting streamed response after error: client disconnected"* — **29 events, 4 users, still firing 8h after it appeared**. First seen `13:40:11Z`, 3½ minutes after the first `1.11.4` pod started. All on `GET /:db/_all_docs`, HTTP **200**, clients Mobile Safari / iOS, 28 of 29 on one tenant.

`forwardToResponse` rejects with `new Error('client disconnected')` when the client goes away mid-stream, and `streamFiltered` logs that via `logger.warn`. `SentryLogger` forwards every `warn` to Sentry, so ordinary client behaviour — navigating away, backgrounding a tab, a mobile network dropping — is filed as a server-side issue. `_all_docs` is by far the slowest endpoint (p95 1652ms over 10.4k requests), so this will keep growing.

The same line also interpolated `error.message` into the message text, so **every distinct downstream error becomes its own Sentry issue** — the fragmentation #294 and #302 have been removing elsewhere. It had already split: [REPLICATION-BACKEND-BG](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-BG) is a second id for the same event.

**Change:** a dedicated `ClientDisconnectedError` so the two cases can be told apart without matching on message text. Disconnects log at `debug` (not forwarded to Sentry); genuine stream faults keep a *constant* message with the detail attached as structured context, which is the contract `SentryLogger` already documents and what `document-changes.service.ts` already does.

## 2. `ENOTFOUND` not classified as transient

`isLikelyTransientError` covers `econnrefused`, `econnreset`, `etimedout`, `socket hang up` … but not DNS resolution failures. In a cluster, `ENOTFOUND`/`EAI_AGAIN` is precisely how a dependency being restarted or rescheduled surfaces: the name stops resolving for a few seconds, then resolves again — transient in exactly the sense `ECONNREFUSED` already is.

The cost showed up in yesterday's rolling redeployment. [REPLICATION-BACKEND-AW](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-AW) logged **189 events across all 40 production tenants** between `13:32:48Z` and `13:57:22Z`:

```
release breakdown:  1.10.10 → 189 events     1.11.4 → 0 events
per tenant:         exactly 3 failures, clustered in ~7 seconds
tenant order:       sequential — amsoc → asa → berami → biffyberlin → camen → …
since rollout:      0 events in the 9 hours after it completed
```

Every event on the outgoing release, none on the incoming one: each tenant's departing pod briefly lost DNS to its CouchDB while that tenant's database service was recycled, and the feed's exponential backoff recovered it. `isTransient` was reported as `false` on those events, which is why they went to `warn` and into Sentry rather than staying at info level.

**Change:** add `enotfound` and `eai_again` to the transient markers. A *sustained* DNS outage is unaffected — `DocumentChangesService` still escalates to `logger.error("SUSTAINED OUTAGE: …")` once the backoff saturates, which never happened here (`failureCount` peaked at 4).

## Notes for review

- Behaviour toward clients is unchanged: `res.destroy()` on an aborted stream, rethrow when nothing has been sent yet. Only what gets logged, and at which level, changes.
- The judgement call worth a second opinion is whether losing the per-disconnect Sentry record is acceptable. It stays in the container logs at debug; what disappears is one Sentry issue per client network event.
- Existing controller tests pass unchanged — the mid-transfer failure test exercises the non-disconnect branch and still asserts `res.destroy()`. Added classification tests for both DNS markers.
- I could not run the suite in this environment; CI is the check.

Related: [ndb-core#4188](https://github.com/Aam-Digital/ndb-core/issues/4188) tracks the same "one fault, many issue ids" pattern on the frontend side.

https://claude.ai/code/session_01WWRV12BxnwWHbqp41pP1mg

---
_Generated by [Claude Code](https://claude.ai/code/session_01WWRV12BxnwWHbqp41pP1mg)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of client disconnects during bulk document operations.
  - Reduced unnecessary warning logs when a client closes a connection before completion.
  - Improved retry classification for temporary DNS resolution failures.
- **Diagnostics**
  - Stream failures are now logged with clearer severity and error details.

### DIFF
--- a/src/common/client-disconnected.error.ts
+++ b/src/common/client-disconnected.error.ts
@@ -1,0 +1,15 @@
+/**
+ * Signals that the client closed the connection before the response was
+ * finished, so the remaining work was abandoned.
+ *
+ * This is ordinary client behaviour — navigating away, backgrounding a tab, a
+ * mobile network dropping — and not a fault of this service. It is a dedicated
+ * type rather than a plain `Error` so the handling code can recognise it
+ * without matching on message text.
+ */
+export class ClientDisconnectedError extends Error {
+  constructor() {
+    super('client disconnected');
+    this.name = 'ClientDisconnectedError';
+  }
+}
--- a/src/common/http-error-classification.spec.ts
+++ b/src/common/http-error-classification.spec.ts
@@ -51,6 +51,21 @@ describe('http-error-classification', () => {
       expect(isLikelyTransientError(new Error('socket hang up'))).toBe(true);
     });
 
+    it('treats DNS resolution failures as transient', () => {
+      // a dependency being restarted or rescheduled stops resolving for a few
+      // seconds; this is how that surfaces, and it recovers on its own
+      expect(
+        isLikelyTransientError(
+          new Error('getaddrinfo ENOTFOUND some-tenant-database'),
+        ),
+      ).toBe(true);
+      expect(
+        isLikelyTransientError(
+          new Error('getaddrinfo EAI_AGAIN some-tenant-database'),
+        ),
+      ).toBe(true);
+    });
+
     it('returns false for non-transient errors', () => {
       expect(
         isLikelyTransientError(
--- a/src/common/http-error-classification.ts
+++ b/src/common/http-error-classification.ts
@@ -15,6 +15,13 @@ const LIKELY_TRANSIENT_ERROR_MARKERS = [
   'bad gateway',
   'gateway timeout',
   'temporarily unavailable',
+  // DNS resolution failures. In a cluster these are how a dependency being
+  // restarted, rescheduled or recreated surfaces to us: the service name stops
+  // resolving for a few seconds and then resolves again. That is transient in
+  // exactly the same sense as ECONNREFUSED, and treating it as a hard fault
+  // means every rolling redeployment reports a warning per retry per tenant.
+  'enotfound',
+  'eai_again',
 ];
 
 export function isAuthHttpError(err: unknown): err is HttpException {
--- a/src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
+++ b/src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
@@ -16,6 +16,7 @@ import { pipeline } from 'stream/promises';
 import { from, Observable } from 'rxjs';
 import { CombinedAuthGuard } from '../../../auth/guards/combined-auth/combined-auth.guard';
 import { User } from '../../../auth/user.decorator';
+import { ClientDisconnectedError } from '../../../common/client-disconnected.error';
 import {
   JsonArrayFilterTransform,
   jsonTokenParser,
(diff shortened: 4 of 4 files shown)


sleidig:
Monitoring run, 11 Aug 2026. Flagging that the issue this PR fixes has grown substantially in the six days since it was opened — the numbers in the description are now well out of date.

[REPLICATION-BACKEND-BF](https://aam-digital.sentry.io/issues/REPLICATION-BACKEND-BF):

| | at time of writing (5 Aug) | now (11 Aug) |
|---|---|---|
| events | 29 | **130 / 24h · 587 lifetime** |
| users affected | 4 | **39** |
| instances | 1 (28 of 29) | **6** |
| substatus | — | **`escalating`** |

The prediction in the description — "`_all_docs` is by far the slowest endpoint … so this will keep growing" — held. Current figures: `GET /:db/_all_docs` at 10,900 requests over 24h, p50 96ms, **p95 1692ms**, still the outlier by a wide margin (`_changes` is p95 115ms over 58,500 requests). The 130 disconnects are ~1.2% of `_all_docs` requests, which is a plausible rate of clients navigating away or dropping network mid-stream rather than anything anomalous.

What has changed qualitatively is the spread: this was 28-of-29 events on a single instance when written, and is now distributed across six, one of which is a dev host. That removes the "maybe it's one tenant with a bad network" reading and supports the PR's framing that this is ordinary client behaviour being recorded as a server fault.

I re-derived the mechanism from source independently this run and reached the same conclusion the description states: `forwardToResponse()`'s `onClose` handler rejects with `new Error('client disconnected')` when the socket closes before `writableFinished`, and `streamFiltered()`'s catch reports it via `logger.warn` once headers are already sent. HTTP status is 200 and the client is gone, so there is no user impact — the cost is purely signal quality, and it is currently the loudest issue in the replication-backend project.

No change requested and nothing rebased — this is a status note so the growth is visible when the PR is picked up. The judgement call the description flags for a second opinion (whether losing the per-disconnect Sentry record is acceptable) is still the open question and still a human decision; this routine is not going to answer it.

One adjacent finding from the same run, in case it is useful context for reviewing the Sentry-noise reasoning here: [#316](https://github.com/Aam-Digital/replication-backend/issues/316) documents a single identity-provider blip producing 16 identical `logger.error` events from one trace, because `getAccessToken()` has no in-flight coalescing. Different code path, same underlying theme as this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

sleidig (commented):
Looks good

PR #321, opened 2026-08-17
feat: pagination works correctly with filtered docs due to permissions

TheSlimvReal:
A request on the `_find` endpoint can come with pagination. Similar to the changes feed, the result might be filtered due to permissions. A similar approach to the changes feed is used to reduce the amount of requests needed to provide a full response.

## Summary by CodeRabbit

* **New Features**
  * Added incremental streaming for JSON array responses, allowing large result sets to be delivered progressively.
  * Limited document searches now filter results while maintaining requested limits and continuing across pages when necessary.
  * Changes and document search responses now support improved backpressure and connection handling.

* **Bug Fixes**
  * Improved response completion and error handling when clients disconnect or servers return incomplete result batches.

### DIFF
--- a/src/common/json-array-response-stream.ts
+++ b/src/common/json-array-response-stream.ts
@@ -1,0 +1,125 @@
+import { Response } from 'express';
+
+/**
+ * Streams a JSON response whose primary payload is one incrementally-written
+ * array field. The response envelope is opened lazily — only after the first
+ * successful upstream batch — so that errors raised before any data is ready
+ * still produce a proper HTTP error status instead of a truncated stream.
+ *
+ * Subclasses provide the array-field name via the constructor and expose
+ * domain-specific `write*` / `finish` methods that delegate to
+ * {@link writeItems} and {@link closeWith}.
+ */
+export abstract class JsonArrayResponseStream {
+  protected written = 0;
+  private opened = false;
+
+  protected constructor(
+    private readonly res: Response,
+    private readonly openingChunk: string,
+  ) {}
+
+  /** Whether the client has disconnected. */
+  get isClosed(): boolean {
+    return this.res.destroyed;
+  }
+
+  /** Number of docs written to the response so far. */
+  get docsWritten(): number {
+    return this.written;
+  }
+
+  /**
+   * Serialize and stream one batch of items into the array, opening the
+   * envelope on the first call. Flushes after each batch so compression
+   * middleware does not buffer the data until the response ends.
+   */
+  protected async writeItems<T>(
+    items: T[],
+    serialize: (item: T) => unknown,
+  ): Promise<void> {
+    await this.open();
+    for (const item of items) {
+      await this.writeChunk(
+        (this.written > 0 ? ',' : '') + JSON.stringify(serialize(item)),
+      );
+      this.written++;
+    }
+    this.flush();
+  }
+
+  /**
+   * Close the array, append the envelope tail and end the response.
+   * Ensures the envelope is opened even when no items were written.
+   */
+  protected async closeWith(closingChunk: string): Promise<void> {
+    await this.open();
+    await this.writeChunk(closingChunk);
+    this.res.end();
+  }
+
+  /**
+   * Send the response headers and open the JSON envelope (idempotent).
+   *
+   * Deliberately called only after the first upstream batch succeeds, so that
+   * upstream errors (e.g. unknown db) still yield a proper error status instead
+   * of a truncated stream.
+   */
+  private async open(): Promise<void> {
+    if (this.opened) return;
+    this.opened = true;
+    this.res.status(200);
+    this.res.setHeader('content-type', 'application/json');
+    await this.writeChunk(this.openingChunk);
+  }
+
+  /**
+   * Hand the buffered bytes to the client right away.
+   *
+   * Without this the compression middleware keeps a batch in its gzip buffer
+   * until enough data has accumulated, so nothing reaches the client until the
+   * response ends and the incremental writing above has no effect for the
+   * (compressing) clients that actually sync.
+   */
+  private flush(): void {
+    (this.res as Response & { flush?: () => void }).flush?.();
+  }
+
+  /**
+   * Write a chunk to the response, awaiting the drain event when the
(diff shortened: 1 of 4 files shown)


src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
```diff
@@ -97,14 +130,78 @@ export class BulkDocEndpointsController {
...
+        stream,
+      );
+      await stream.finish(bookmark);
+    } catch (error) {
+      if (!res.headersSent) throw error;
+      this.logger.warn('aborting streamed _find response after error', {
+        error: error instanceof Error ? error.message : String(error),
+      });
+      res.destroy();
+    }
```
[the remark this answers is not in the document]
TheSlimvReal: I would leave it the way it is now, especially the time limit seem dangerous because the frontend needs to know if the full page can be created or not.

src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
```diff
@@ -97,14 +130,78 @@ export class BulkDocEndpointsController {
...
+        }),
+      );
+
+      const permitted = response.docs.filter(isPermitted).slice(0, remaining);
+      await stream.writeDocs(permitted);
+
+      bookmark = response.bookmark;
+
+      if (response.docs.length < internalLimit) break;
+      if (!response.bookmark) break;
```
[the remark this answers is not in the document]
TheSlimvReal: Bookmarks are not send back as they might not be correct

src/common/json-array-response-stream.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Streams a JSON response whose primary payload is one incrementally-written
+ * array field. The response envelope is opened lazily — only after the first
+ * successful upstream batch — so that errors raised before any data is ready
+ * still produce a proper HTTP error status instead of a truncated stream.
+ *
+ * Subclasses provide the array-field name via the constructor and expose
+ * domain-specific `write*` / `finish` methods that delegate to
+ * {@link writeItems} and {@link closeWith}.
+ */
+export abstract class JsonArrayResponseStream {
```
sleidig, line 13: we recently fixed a bug with special characters getting broken by our streaming in aam-services: https://github.com/Aam-Digital/aam-services/pull/181

Let's do a critical agent review if anything similar could possibly happen here. Looks safe to me, but let's be double sure 😊

src/common/json-array-response-stream.ts
```diff
@@ -0,0 +1,125 @@
...
+        cleanup();
+        resolve();
+      };
+      // socket error or close while back-pressured: settle the promise so the
+      // request handler never hangs (which would leak a CouchDB keep-alive
+      // socket). 'error' is required — without it a socket error that does not
+      // also emit 'close' would leave this promise pending forever.
+      const onFailure = () => {
+        cleanup();
+        reject(new Error('client disconnected'));
```
sleidig, line 118: are we still catching "Client disconnected" and filtering it from sentry logging? (this used to spam our sentry until adding filters recently)

sleidig (approved):
Nice. I like the generic implementation. Looks good to me overall.

src/common/json-array-response-stream.ts
TheSlimvReal: Claude analysis:

The same bug does not affect this implementation. Here's why:

The Kotlin bug was stateless per-chunk UTF-8 decoding — each raw byte buffer was independently decoded as a complete UTF-8 document, so a multi-byte character (e.g., Devanagari's E0 A4 BE) split across a read boundary would produce replacement characters from both halves.

This codebase has two data paths, and neither has that problem:

Path 1 — _changes and paginated _find (via ChangesController/streamPermittedFindDocs)
The HTTP client (couchdbService.get()/post()) fully buffers and parses the CouchDB JSON response into a JavaScript object before it reaches this code. There's no per-chunk decoding happening at all; the values are already proper JS strings by the time JSON.stringify() is called in writeItems.

Path 2 — _bulk_get, _all_docs, unlimited _find (via streamFiltered)
This path does handle raw bytes: couchdbService.postStream() returns a Node.js Readable, piped through jsonTokenParser() (i.e., stream-json's parserStream). But stream-json uses Node.js's built-in StringDecoder under the hood, which is explicitly designed to be stateful across chunks — it holds back an incomplete multi-byte sequence at the end of a chunk and prepends it to the next one. That's the standard Node.js fix for exactly this class of bug. After parsing, values are assembled into JS objects and re-serialized with JSON.stringify(), so the round-trip through V8's string model handles all Unicode correctly.

The Kotlin code was doing what Node.js's StringDecoder was designed to prevent. You're safe here.

src/common/json-array-response-stream.ts
TheSlimvReal: Its the exact same code as previously used in the `ChangesController` so I don't know how it was previously filtered but it shouldn't have changed anything

src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts
```diff
@@ -60,8 +60,8 @@ class FindResponseStream extends JsonArrayResponseStream {
   }
 
   /** Append the bookmark envelope field and end the response. */
```
sleidig: @TheSlimvReal the comment is still talking about the bookmark.

What is the bookmark actually for and why did we remove it?

TheSlimvReal: The bookmark allows to quickly fetch the next batch of data, like a cursor. It does not really make sense to return it because when we are e.g. requesting 500 docs but only need 300 to fulfil the request of the user, the other 200 are discarded. The bookmark still points to after the 500 docs, so when the app would now use this bookmark in a later request, it would skip the 200 discarded docs. Of course this could be fixed but we don't use the bookmark anywhere so I just removed it to make sure that nobody accidentally uses while this problem exists.

PR #328, opened 2026-08-25
feat(permissions): managed baseline rules for public form visitors

sleidig:
Anonymous visitors of a public form have no system-managed permission baseline at all.

`RulesService.getRulesForUser()` short-circuits with `if (!user) return this.permission?.[PUBLIC_SECTION_KEY] ?? []`, so `_default` — and with it `MANAGED_DEFAULT_RULES` — never applies to them. Everything an unauthenticated visitor can do is hand-authored per instance.

That matters most for one document, because of a bootstrap recursion: **to know what `_public` allows, the client must first read `Config:Permissions`.** Where an instance doesn't happen to grant that, `DocumentController.getDocument` answers 401 (`UnauthorizedException('unauthorized', 'User is not permitted')` — a rules denial, not an auth failure, since `CombinedAuthGuard` lets anonymous requests through), the frontend cannot load its own rule set, and falls back to allowing everything locally. `CONFIG_ENTITY` and `SiteSettings` are what a public form is rendered from.

### What this adds

`MANAGED_PUBLIC_RULES` — read on `Config:Permissions` + `Config:CONFIG_ENTITY` (scoped by `_id`, like the existing managed rule) and read on `SiteSettings`.

`mergeManagedRules()` generalises the existing merge; `mergeManagedPublic()` applies the policy:

- returns `undefined` when there is no `_public` section — **the section is never created**, only extended, so an instance without public forms is never granted anonymous access;
- adds the managed rules exactly while the section holds at least one admin-authored rule;
- **strips them again** when the last admin rule is removed, so an instance that stops using public forms stops granting the access.

Self-healing in both directions: `ensureManagedDefaults` runs on every change to the permission document, so an instance that later adds its first `_public` rule picks up the baseline on that same write — no migration step.

`writeManagedDefaults` now merges both sections in one write, including the legacy `public` → `_public` key migration the `default` section already had.

### Note on disclosure

This makes the permission document readable by anyone holding a public-form URL on instances that use them: role names, entity types, rule conditions. No user data. It is also *more* precise than the status quo — all `Config` docs share the CASL subject `Config`, so instances that wrote a broad `read: Config` into `_public` already expose it, while this rule is `_id`-scoped.

### Effect on the frontend

The ordinary path then works end to end — `GET /:db` and `GET /:db/_changes` already permit anonymous callers, so the document rules were the only blocker. The `_public` rules get applied client-side, which activates the existing `no_permissions` state in `public-form.component.ts` where the public role lacks create permission, instead of the visitor filling in a form that fails on submit.

Related: Aam-Digital/ndb-core#4315, Aam-Digital/ndb-core#4198

## Summary by CodeRabbit

* **New Features**
  * Added managed anonymous-access permissions for public forms, including required configuration and site-settings access.
  * Public permission sections are now automatically enriched and legacy public sections are migrated while preserving custom rules.
  * Permission updates now report removed managed rules and persist changes across applicable sections.

* **Bug Fixes**
  * Prevented unnecessary creation or updates of public permission sections when no public form exists.
  * Preserved explicitly empty sections and administrator-defined permissions during rule updates.

* **Tests**
  * Expanded coverage for permission merging, migration, idempotence, and write-back behavior.

### DIFF
--- a/src/permissions/rules/default-permissions.ts
+++ b/src/permissions/rules/default-permissions.ts
@@ -60,22 +60,62 @@ export const MANAGED_DEFAULT_RULES: DocumentRule[] = [
 ];
 
 /**
- * Merge the managed default rules into an existing `default` rules section.
+ * Baseline rules for unauthenticated visitors of a public form.
+ *
+ * `_public` is the `_default` equivalent for anonymous traffic: the `_default`
+ * section is only merged for authenticated accounts (see
+ * `RulesService.getRulesForUser`), so unauthenticated clients need their own
+ * managed baseline.
+ *
+ * Unlike {@link MANAGED_DEFAULT_RULES} these are only written into a `_public`
+ * section that an admin already uses (see {@link mergeManagedPublic}) - an
+ * instance without public forms grants anonymous visitors nothing.
+ */
+export const MANAGED_PUBLIC_RULES: DocumentRule[] = [
+  {
+    // Config:Permissions is what makes the other rules enforceable: a client
+    // that cannot read its own rule set has nothing to enforce and falls back
+    // to allowing everything locally. CONFIG_ENTITY holds the entity/field
+    // definitions a public form is rendered from.
+    action: 'read',
+    subject: 'Config',
+    conditions: {
+      _id: { $in: ['Config:CONFIG_ENTITY', 'Config:Permissions'] },
+    },
+    reason: `${SYSTEM_DEFAULT_MARKER} public form config read access`,
+  },
+  {
+    // branding (logo, colors) of the page the form is displayed on
+    action: 'read',
+    subject: 'SiteSettings',
+    reason: `${SYSTEM_DEFAULT_MARKER} public form site settings read access`,
+  },
+];
+
+/** Outcome of merging a managed rule set into one section of the config. */
+export interface ManagedRulesMerge {
+  merged: DocumentRule[];
+  changed: boolean;
+  dropped: DocumentRule[];
+  /** the section's admin-authored rules, i.e. everything not system-managed */
+  adminRules: DocumentRule[];
+}
+
+/**
+ * Merge a managed rule set into an existing rules section.
  * Managed rules are prepended and any previously written system-default rules
  * are replaced by the current managed set, so updated backend versions
  * refresh their own rules while admin-authored rules stay untouched.
  *
- * A non-array `currentDefault` (e.g. `null` from a malformed document) is
+ * A non-array `currentSection` (e.g. `null` from a malformed document) is
  * treated the same as an empty section instead of throwing, so that a
- * malformed `default` section gets actively healed rather than crashing the
- * caller.
+ * malformed section gets actively healed rather than crashing the caller.
  */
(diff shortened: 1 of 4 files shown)


sleidig:
Addressed the review, plus one thing the CodeRabbit walkthrough surfaced.

**The branch was based on the wrong commit.** It picked up `475509e fix(logging): only warn about _changes requests that are actually slow` — which belongs to `fix/changes-slow-warning-threshold`, not `master` — so this PR was reviewing an unrelated `changes.controller.ts` change and was effectively stacked on that branch. Rebased onto `origin/master`; the PR now contains only its own commits. That commit is untouched on its own branch.

**CodeRabbit — canonical `_default` in the JSDoc:** valid, fixed. Rather than only correcting the wording, both sections are now read through the existing `normalizeSectionKeys()`, so the write-back no longer handles each legacy key inline. That removes the duplicated `?? LEGACY_*` / `delete` pairs, makes "the doc carried a legacy key" a single condition, and drops the deprecated-symbol references entirely.

**qlty:**
- `typescript:S1874` (`LEGACY_PUBLIC_KEY` / `public` deprecated, 4 sites) — resolved by the same refactor rather than suppressed. Reading a deprecated key was intentional for migration, but it only needs to happen in one place, which already existed.
- `typescript:S4325` (unnecessary assertion) — replaced the `as ManagedRulesMerge` cast with a guard that throws, so the narrowing is real rather than asserted.
- `qlty:similar-code` in `rules.service.spec.ts` — the three new write-back tests are grouped in a `describe` that shares the fake-timer and changes-feed boilerplate through one `writeBackFor()` helper, so each test is now just its input document and its assertion.

`qlty check --upstream origin/master` is clean apart from one `eslint` "File ignored because no matching configuration was supplied" on a spec file, which is a local tooling artefact and reproduces with the repo's own `npx eslint`.

76 permissions tests pass. No behaviour change in this round: `normalizeSectionKeys()` applies the same precedence the inline handling did (current key wins over legacy) and returns the config unchanged when there is nothing to migrate.

Abhinegi2:
app working as expected; now doing code review

src/permissions/rules/default-permissions.ts
```diff
@@ -60,22 +60,62 @@ export const MANAGED_DEFAULT_RULES: DocumentRule[] = [
...
- * caller.
+ * malformed section gets actively healed rather than crashing the caller.
  */
-export function mergeManagedDefaults(currentDefault: DocumentRule[] = []): {
-  merged: DocumentRule[];
-  changed: boolean;
-  dropped: DocumentRule[];
-} {
-  const current = Array.isArray(currentDefault) ? currentDefault : [];
+export function mergeManagedRules(
```
Abhinegi2: It seems we’re only using this function in this file, so we don’t need to export it

Abhinegi2 (changes requested):
code also looks good to me. I just have one small cleanup suggestion

src/permissions/rules/default-permissions.ts
sleidig:
```suggestion
function mergeManagedRules(
```

Abhinegi2 (approved):

PR #339, opened 2026-09-02
feat: added fields support for find request

TheSlimvReal:


TheSlimvReal:
See https://github.com/Aam-Digital/ndb-core/issues/4314#issuecomment-5581744729

PR #340, opened 2026-09-04
chore: remove the legacy permission section keys

### LINKED ISSUE #295: Remove backwards compatibility for legacy permission section keys
The reserved sections of the `Config:Permissions` document were renamed `default` -> `_default` and `public` -> `_public`, so that they can no longer be shadowed by a realm role of the same name.

Backend and frontend still read the legacy spellings, and the backend moves a legacy `default` section onto `_default` when it writes the managed defaults back. That backwards compatibility exists only so code can ship before every document has been migrated, and it should be removed in the next major version.

### To remove

replication-backend:
- https://github.com/Aam-Digital/replication-backend/pull/289
- `LEGACY_DEFAULT_KEY`, `LEGACY_PUBLIC_KEY` and `LEGACY_SECTION_KEYS` in `src/permissions/rules/permission.ts`
- the deprecated `public` / `default` members of `RulesConfig`, and the legacy entries in `RESERVED_RULE_CONFIG_KEYS`
- `RulesService.normalizeSectionKeys` and its call in `setPermission`
- the legacy branch in `RulesService.writeManagedDefaults` (read `_default` only)

ndb-core:
- the legacy key fallbacks added in Aam-Digital/ndb-core#4168

### Precondition

Every instance has to run the one-off config migration that renames the keys in its permission document first, otherwise its `default` / `public` rules stop being applied.

Abhinegi2:
- closes #295

Part of #295 (backend half). The ndb-core half is in Aam-Digital/ndb-core#4354.

The reserved sections of `Config:Permissions` were renamed `default` -> `_default` and `public` -> `_public`. The non-prefixed spellings stayed readable only so code could ship before every document had been migrated. All instances now have `_default` / `_public` in their CouchDB, so the compatibility layer can go.

## Removed

- `LEGACY_DEFAULT_KEY`, `LEGACY_PUBLIC_KEY` and `LEGACY_SECTION_KEYS` in `src/permissions/rules/permission.ts`
- the deprecated `public` / `default` members of `RulesConfig`, and their entries in `RESERVED_RULE_CONFIG_KEYS`
- `RulesService.normalizeSectionKeys` and its call in `setPermission`
- the legacy branch of `RulesService.writeManagedDefaults` — it now reads `_default` / `_public` only, and no longer rewrites a document just to migrate a section key

Behaviour for an already-migrated document is unchanged; only the migration path is gone.

## Precondition

**An instance that has not run the key-rename migration loses its shared `default` / `public` rules with this change** — they are simply no longer read. Both one-off migrations (`oneoff-20260724-permissions-key-rename`, then `oneoff-20260804-permissions-key-legacy-cleanup`) must have run everywhere before this is merged.

Draft until that is confirmed, and until the matching ndb-core PR is ready — the two should be released together so no deployed frontend is left reading a spelling the backend no longer writes.

## Testing

- `npm test` — 290 passed, 32 suites
- `npm run build` — clean

Tests that only covered the legacy-to-current migration were removed; the ones that used `default:` / `public:` merely as a key name now use the prefixed spelling.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed support for legacy `default` and `public` permission section keys.
  * Permission configurations must use the current `_default` and `_public` section keys.
  * Legacy section-key migration and automatic normalization are no longer performed.

### DIFF
--- a/src/permissions/rules/permission.ts
+++ b/src/permissions/rules/permission.ts
@@ -7,54 +7,32 @@ export const ADMIN_APP_ROLE = 'admin_app';
 /**
  * Section keys in {@link RulesConfig} that carry special semantics instead of
  * mapping a user role. The underscore prefix marks them as internal so they
- * cannot collide with a realm role name. Legacy (non-prefixed) spellings are
- * still read for backward compatibility until all documents are migrated.
+ * cannot collide with a realm role name.
  */
 export const DEFAULT_SECTION_KEY = '_default';
 export const PUBLIC_SECTION_KEY = '_public';
-/** @deprecated use {@link DEFAULT_SECTION_KEY}; only read for migration */
-export const LEGACY_DEFAULT_KEY = 'default';
-/** @deprecated use {@link PUBLIC_SECTION_KEY}; only read for migration */
-export const LEGACY_PUBLIC_KEY = 'public';
-
-/**
- * Legacy section key -> current section key. A config that still uses a legacy
- * key is normalized onto the current one when it is loaded, so only migration
- * code needs to know about the old spellings.
- *
- * @deprecated remove together with the legacy keys in the next major version
- */
-export const LEGACY_SECTION_KEYS: Record<string, string> = {
-  [LEGACY_DEFAULT_KEY]: DEFAULT_SECTION_KEY,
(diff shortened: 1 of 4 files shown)


sleidig (approved):
Looks good to me 👍

sleidig:
@coderabbitai review

PR #351, opened 2026-09-14
feat: Added support for paginated view queries

TheSlimvReal:
Currently, only the `_find` endpoint supported pagination. This now also introduces pagination for the `_design/../_view/...` endpoint. This is done by modifying the returned offset key.

Approach for a client to use pagination:

1. Send request with `startkey`, `endkey` and `limit` to get the first page
2. Save `offset` from response
3. To fetch next page, omit `startkey` and send request with `skip = limit + offset`, `limit` and `endkey` as before
4. Repeat 2 & 3 for further pages

This approach allows efficient pagination with permissions enabled.

## Summary by CodeRabbit

- **New Features**
  - Added support for paginated view queries when requesting documents with a valid result limit.
  - Large permitted result sets can now be streamed progressively instead of waiting for the full response.
  - Responses include an adjusted offset to support paging through filtered results.

- **Bug Fixes**
  - Improved handling of permission-filtered view results and exhausted result sets.
  - Added safer handling for interrupted client connections and errors during streaming.


TheSlimvReal:
@coderabbitai review

src/restricted-endpoints/document/design-document.controller.ts
```diff
@@ -113,36 +149,153 @@ export class DesignDocumentController {
...
-        if (isDeletedRow) {
-          return true;
-        }
+      queryParams.include_docs === 'true' || queryParams.include_docs === true;
+    const limit = this.parseInt(queryParams.limit, 1);
+    const rowFilter = this.viewRowFilter(user);
+
+    if (!includeDocs || limit === undefined) {
+      const result = await firstValueFrom(
+        this.couchdbService.get<ViewResponse>(db, viewPath, queryParams),
```
sleidig: why are we not streaming results here?

